PR #48, opened 2023-07-11
[Docs] Updated 'Miscellaneous Examples' Section

aku1310:
Keywords already in use: `$id`, `$schema`, `title`, `properties`, `description`, `type`, `minimum`, `maximum`, `required`, `$defs`, and `$ref`.

Keywords added: `enum`, `pattern`, `dependencies`, `if`, `then`, `else`.

I have added some examples which showcased the use of other keywords and structures, and were relatively shorter to be added in the miscellaneous section.

I feel like the other examples which use vocabulary for specific application formats, hypermedia, etc. can be made longer to properly show their usage using a specific use case as well. 

Please let me know what you think about this.


pages/learn/miscellaneous-examples.md
```diff
@@ -157,3 +174,243 @@ We also introduce the following with this example:
...
+    "postalCode": "10001"
+  },
+  "hobbies": ["reading", "running"]
+}
+```
+
+The provided data conforms to the schema by including values for the required properties and ensuring the `age` is an integer greater than or equal to zero. The `address` object contains all the necessary properties, and the `hobbies` property is an array of strings.
+
+
+## Conditional validation with dependencies
```
gregsdennis, line 308: `dependencies` has been split into `dependentRequired` for property requirements and `dependentSchemas` for schemas requirements.  It would be good to have examples of both.

`dependencies` is now deprecated.

pages/learn/miscellaneous-examples.md
```diff
@@ -157,3 +174,243 @@ We also introduce the following with this example:
...
+  "properties": {
+    "isMember": {
+      "type": "boolean"
+    },
+    "membershipNumber": {
+      "type": "string",
+      "maxLength": 10,
+      "minLength": 10,
+      "dependencies": {
+        "isMember": true
```
gregsdennis, line 327: > ... only when the `isMember` property is set to true.

This isn't a correct usage of `dependencies`.  `dependencies doesn't support what you're claiming.  You'd need the proposed `propertyDependencies` keyword to match on property values like this.

`dependentProperties` contains lists of properties that are required when another property is present.

```json
{
  "dependentProperties": {
    "foo": [ "bar" ]
  }
}
```

This requires that `bar` be present when `foo` is present, but it doesn't require that `foo` or `bar` have any particular value, just that they exist.  Think of it as a conditional `required`.

`dependentSchemas` contains a subschema that the instance must also match if a property is present.

```json
{
  "dependentSchemas": {
    "foo": { "minProperties": 7 }
  }
}
```

This requires that the instance have 7 properties if `foo` is present.

pages/learn/miscellaneous-examples.md
```diff
@@ -157,3 +174,243 @@ We also introduce the following with this example:
...
+    "membershipNumber": {
+      "type": "string"
+    }
+  },
+  "if": {
+    "properties": {
+      "isMember": {
+        "const": true
+      }
+    }
```
gregsdennis, line 377: This needs a `required: [ isMember ]` to work.  Otherwise it'll trigger when `isMember` is absent.

pages/learn/miscellaneous-examples.md
```diff
@@ -157,3 +174,243 @@ We also introduce the following with this example:
...
+  },
+  "else": {
+    "properties": {
+      "membershipNumber": {
+        "type": "string"
+      }
+    }
+  }
```
gregsdennis, line 394: This is redundant since it doesn't add any constraints not already defined in the root schema.

gregsdennis (changes requested):
Thanks for submitting!  Some things to take care of.

aku1310:
> Thanks for submitting! Some things to take care of.

Thank you for the feedback! 
I'll make the suggested changes and update the PR soon.

pages/learn/miscellaneous-examples.md
gregsdennis: @aku1310 I mistyped on this at first and put `dependentProperties` when I meant `dependentRequired`.  I've corrected the above but wanted to explicitly say something in case you've already started your changes.

benjagm (changes requested):
I love what you have done to provide a better introduction to each example. That was exactly what I was referring to when I asked to provide a more consistent format for examples.

Regarding the code and keywords thanks a lot @gregsdennis for the comments here. Those are the changes we should include. If you need additional inputs feel free to reach out to @gregsdennis or me and we'll help you. 

Nice job overall! You are taking our docs to the next level!

aku1310:
Hi @gregsdennis @benjagm, 

I have updated the PR with the suggested changes. Please let me know what you think about this. 😊

Thank you!

pages/learn/miscellaneous-examples.md
```diff
@@ -157,3 +173,291 @@ We also introduce the following with this example:
...
+  },
+  "dependentRequired": {
+    "foo": ["bar"]
+  }
+}
+```
+
+**Data**
+
+```json
```
gregsdennis, line 332: It might be good to show an invalid instance where `bar` is missing as well as a valid instance where both `foo` and `bar` are missing, and explanations why.

pages/learn/miscellaneous-examples.md
```diff
@@ -157,3 +173,291 @@ We also introduce the following with this example:
...
+        }
+      }
+    }
+  }
+}
+```
+
+**Data**
+
+```json
```
gregsdennis, line 380: It might be good to show an invalid instance where `propertiesCount` is not an integer as well as a valid instance where both `foo` is missing and `propertiesCount` is not an integer, and explanations why.

pages/learn/miscellaneous-examples.md
```diff
@@ -157,3 +173,291 @@ We also introduce the following with this example:
...
+
+
+## Conditional validation with if-else
+
+In this schema, we have two properties: `isMember` and `membershipNumber`. The conditional validation is based on the value of the `isMember` property. The validation keywords [if](https://json-schema.org/draft/2020-12/json-schema-core.html#section-10.2.2.1), [then](https://json-schema.org/draft/2020-12/json-schema-core.html#name-then), and [else](https://json-schema.org/draft/2020-12/json-schema-core.html#name-else).
+
+Here's how the validation works in this example:
+
+If the value of `isMember` is true:
+* The then block is applied, which specifies that the `membershipNumber` property should be a string with a minimum length of 10 and a maximum length of 10.
```
gregsdennis, line 397: Format for `then`, please.

pages/learn/miscellaneous-examples.md
```diff
@@ -157,3 +173,291 @@ We also introduce the following with this example:
...
+
+If the value of `isMember` is anything other than true:
+* The else block is applied, which specifies that the `membershipNumber` property can be any string.
```
gregsdennis, line 400: Format for `else`, please.

gregsdennis (commented):
Very good edits. I've left a few comments.

aku1310:
> Very good edits. I've left a few comments.

Thank you!! I'll make the changes and update the PR soon.

pages/learn/miscellaneous-examples.md
aku1310: I mistakenly set the `propertiesCount` type as an integer under the `dependentSchemas` again. 😅 
The type is an integer only. However, for the invalid instance, I can set it to be less than 7 instead of the minimum condition of 7. Similarly, for the valid instance where `foo` is missing, I can also set it to be less than 7 since the minimum value should be 0 in that case. 

Please let me know if this works.

gregsdennis: Ah, yes. I missed the integer declaration up higher.

This sounds good. Thanks.

aku1310: Yes, okay! Will just do it.

pages/learn/miscellaneous-examples.md
```diff
@@ -157,3 +173,328 @@ We also introduce the following with this example:
...
+    "foo": ["bar"]
+  }
+}
+```
+
+**Data**
+
+```json
+{
+  "validity": "valid",
```
gregsdennis, line 337: I don't know that I'd have included validity as part of the data, but it works.

I'd like to get someone else's opinion on this.

gregsdennis (approved):
I'm happy with this.  I'd like to get a second opinion on one thing.

pages/learn/miscellaneous-examples.md
jdesrosiers: I agree. I think it's confusing to have "validity" in the data.

aku1310: Okay, I'll remove that and update the PR.

gregsdennis: Thanks.  Happy with the change.

benjagm (approved):

benjagm (changes requested):
Everything looks great. The only thing I noticed is this page will benefit from providing an index at the beginning to better help users to understand what to expect. Something like:

In this page you will find examples illustrating different uses cases to help you get the most out of your JSON Schemas, including:

* A typical minimum schema 
* Describing geographical coordinates
* Arrays of things
* Enumerated values
* Regular expression pattern
* Complex object with nested properties
* Conditional validation with dependentRequired
* Conditional validation with dependentSchemas
* Conditional validation with if-else

aku1310:
> Everything looks great. The only thing I noticed is this page will benefit from providing an index at the beginning to better help users to understand what to expect. Something like start the page this way instead of providing the first example:
> 
> In this page you will find examples illustrating different uses cases to help you get the most out of your JSON Schemas, including:
> 
> * A typical minimum schema
> * Describing geographical coordinates
> * Arrays of things
> * Enumerated values
> * Regular expression pattern
> * Complex object with nested properties
> * Conditional validation with dependentRequired
> * Conditional validation with dependentSchemas
> * Conditional validation with if-else

Thank you! Just updated this. Please let me know if this works 😊

benjagm:
Approved! Did a small change to fix relative urls in the index links.

PR #54, opened 2023-07-25
Self-host past drafts

### LINKED ISSUE #49: Link to self hosted drafts, not IETF
In a [recent discussion](https://github.com/w3c/strategy/issues/108#issuecomment-1622791662), we learned that we can't count on IETF to host expired I-Ds. We currently count on that for hosting past draft releases. The only one we host is the current version. We need to start hosting the others as well.

This change has a few other benefits as well. We have been using I-Ds for our spec, but not the way they were intended to be used, which means that some of the things that normally apply to I-Ds don't apply to us. Most importantly, our drafts effectively don't expire. As long as they are used in the wild, they are relevant. Since we don't use I-Ds properly, linking to IETF causes confusion.

Which leads to the other benefit in that self-hosting may make other specifications feel more comfortable referencing our past specifications. When we point to IETF, the IETF rules about expiration apply, but when we self-host, we can claim different rules that make sense for our situation.

---

It appears that we currently host most of the draft html, but we only use it for the current spec. I don't know why. So, this task will mostly just require changing links. Drafts that don't already have html are 0, 2, 3, and 5.

jdesrosiers:
Resolves #49 

I updated all links to our past drafts to point to a self-hosted version rather than IETF.

For most releases, we already had a copy in the repo, so I just had to change the link.

### Missing copies
For the releases I didn't have a copy for, I got one from the IETF site. Unfortunately, they don't have an HTML copy that's self contained, so I chose to use the pdf version instead. However, for a couple of the really old releases, the only thing I could get my hands on was the txt version. It's probably possible to go back in git history to find html versions in the old website for some of the releases I used pdfs for, but I'm not sure that's worth the effort. Here are releases that don't have an html version.

* draft-zyp-json-schema-00.txt (draft-00)
* draft-zyp-json-schema-02.txt (draft-02)
* draft-zyp-json-schema-03.pdf (draft-03)
* draft-wright-json-schema-00.pdf (draft-05/draft-04)
* draft-wright-json-schema-validation-00.pdf (draft-05/draft-04)
* draft-wright-json-schema-hyperschema-00.pdf (draft-05)
* draft-handrews-json-schema-00.pdf (draft-07 obsolete)
* draft-handrews-json-schema-validation-00.pdf (draft-07 obsolete)

### Relative JSON Pointer
I left Relative JSON Pointer pointing to IETF. I think most of us agree that that spec makes sense to stay on the IETF track.

### Symlinks
I did a little reorganization of how we store past drafts. Documents are now named with their IETF identifier and there are symlinks to the latest version using the common names. For example, draft-07 had two releases, `draft-handrews-json-schema*-00` and `draft-handrews-json-schema*-01`. All of those documents are in the `/draft07` folder and there are symlinks pointing `/draft-07/json-schema-core.html`, `/draft-07/json-schema-validation.html`, and `/draft-07/json-schema-hyperschema.html` to the `*-01` versions.

I didn't to create the symlinks for drafts 00-03 because they're just so old and unused that I didn't think it was necessary. I didn't create symlinks for draft-05 because it's a weird case and I didn't really know what to do with it. Technically, core and validation are just patch updates to draft-04, so `draft-04/json-schema*.html` should point to the draft-05 documents. But, hyperschema is changed in ways that requires that it be considered it's own version. I decided to just leave draft-05 as a distinct release even though it mostly isn't.

### Other cleanup
While looking for IETF links, I came across some IETF copy that doesn't make sense and either removed or updated it. I also noticed we were hosting a pdf for Understanding JSON Schema. That doesn't make sense in the new site, so I removed it.


jdesrosiers:
I'm not even a little worried that there's a legal issue with us moving away from IETF, but this change isn't related to that choice. It's something we need to do regardless because IETF told us that they don't guarantee that they will host expired drafts forever. Self-hosting is the only choice we have or we could loose those documents. 

The only possible legal issue, it's whether or not we can tell people that they can normatively reference our past IETF-published drafts in their spec because we host them and we said it's ok. If it turns out we can't say that, that would be (and currently is) annoying for other specifications, but it wouldn't mean we don't need to, or can't, self-host those documents.

benjagm (approved):
That's a huge effort you putted on this. Thanks a lot Jason!

Looks good to me.

pages/_index.md
```diff
@@ -2,35 +2,36 @@
...
 We are using dates for meta-schemas, which are what implementations should use to determine behavior,
 so we will usually refer to `2020-12` (without the word "draft") on this web site.
 
-See the [Specification page](/specification) for details about naming and numbering.
+See the [Specification Links](/specification-links) page for details about naming and numbering.
 
 ### The Path to Standardization
 
-The JSON Schema project intends to shepherd all three draft series to either: RFC status, the equivalent within another standards body, and/or join a foundation and establish self publication rules.
+The JSON Schema project intends to self-publish the two JSON Schema specifications with the OpenJS Foundation. For the Relative JSON Pointer specification, we will likely pursue an RFC with IETF.
```
gregsdennis, line 14: Henry is actively pursuing the RJP spec with IETF.  Not sure if that means tweaking this.

pages/_index.md
```diff
@@ -2,35 +2,36 @@
...
+remain relevant as long as they continue to be used in production despite what
+the document declares as its expiration.
 
-Internet-Drafts expire after six months, so our goal is to publish often enough to always have a set of unexpired drafts available.  There may be brief gaps as we wrap up each draft and finalize the text.
 </details>
 
 ### Use of the _draft_ designation
 Releases of the JSON schema specification and meta schemas use the _draft_ designation primarily for historical reasons stemming from the relationship of this specification to IETF ([explained here](https://json-schema.org/specification-links#understanding-draft-names-and-numbers)).
-The use of this designation is under review but will continue until this review process completes to avoid changing the designation style multiple times.
+The use of this designation is under review and will not continue in future releases.
```
gregsdennis, line 34: What does it mean for the use of "draft" to be "under review?"  If we've decided that we're not going to have it in the future, doesn't that mean it's no longer under review?

pages/specification-links.md
```diff
@@ -301,8 +299,8 @@ For links to the somewhat more readably formatted versions on this web site, and
 ### 2020-12
 
 - Specifications
-    - Core: [draft-bhutton-json-schema-01](https://tools.ietf.org/html/draft-bhutton-json-schema-01) ([changes](https://tools.ietf.org/html/draft-bhutton-json-schema-01#appendix-G))
-    - Validation: [draft-bhutton-json-schema-validation-01](https://tools.ietf.org/html/draft-bhutton-json-schema-validation-01) ([changes](https://tools.ietf.org/html/draft-bhutton-json-schema-validation-01#appendix-C))
+    - Core: [draft-bhutton-json-schema-01](/draft/2020-12/draft-bhutton-json-schema-01.html) ([changes](/draft/2020-12/draft-bhutton-json-schema-01.html#appendix-G))
+    - Validation: [draft-bhutton-json-schema-validation-01](/draft/2020-12/draft-bhutton-json-schema-validation-01.html) ([changes](/draft/2020-12/draft-bhutton-json-schema-validation-01.html#appendix-C))
```
gregsdennis, line 303: Why are the other entries split with "Core:" on one line and the links on another, but this one is not?  (I think I prefer not split.)

gregsdennis (changes requested):
Just some questions.  Otherwise, looks great!  Thanks.

pages/_index.md
jdesrosiers: I think it's fine. RJP hasn't yet been adopted by an IETF working group. If you want to propose alternate wording, I'd be happy to include it.

pages/_index.md
jdesrosiers: Good point. I guess I was trying not to change things any more than was necessary, but didn't go far enough.

pages/specification-links.md
jdesrosiers: I think they're all supposed to be on the same line. There seems to be a mix. I'll normalize them.

gregsdennis (approved):

PR #59, opened 2023-08-21
clean up alignment, weight, blog filters

OlliesWorld:
fix 
font-weight for search button, 
hover styles on hover are better but not fully fixed
logo alignment with the supported section, 
blog filters added `All` and update date on 1 post to clean up hero img


OlliesWorld:
not needed

PR #62, opened 2023-09-01
[Docs] Updated Examples Section

aku1310:
In **Other Examples**:
- Added more use cases of JSON Schema
- Added a short summary and index at the top

In **Miscellaneous Examples**:
- Removed Geographical Locations example as it is more suited in 'Other Examples' section
- Renamed it to Generic Examples

In **Modelling a File System**:
- Added a short summary before the index

Please let me know what you think about this and any changes required.


benjagm (changes requested):
Nice job. I am providing some small suggestions for better consistency among the different example pages.

We'll need more people reviewing the examples.

aku1310:
> Nice job. I am providing some small suggestions for better consistency among the different example pages.
> 
> We'll need more people reviewing the examples.

Thank you! I've made the suggested changes.

benjagm (approved):
Looks good to me. Nice job!

I'd love to get at least another review.

pages/learn/json-schema-examples.md
```diff
@@ -41,7 +63,126 @@ title: JSON Schema examples
...
+  "region": "State",
+  "postal-code": "12345",
+  "country-name": "Country"
+}
+```
+
+
+## Blog post
+
+A schema representing a blog post, including properties like `title`, `content`, `published date`, `author details`, and `tags`.
```
gregsdennis, line 82: If you're going to use the "speech" form of these properties (e.g. "published date" instead of `publishedDate`), I think it should be without formatting.  Otherwise, just use the actual property names.  Similarly for the other summaries.  We should be consistent as well.

gregsdennis (commented):
These examples look good, and it appears you put some time into them.  However, I would expect each example to showcase some different feature set.

For instance, if you're starting from the top, "blog post" adds nested properties, a feature that "address" doesn't have.  But then "book" just illustrates the same feature set that "address" did.  Does having "book," then, hold any real value?

To me, these all look like basically the same example.  (Maybe that's needed for this page; I'm not sure.)

aku1310:
@gregsdennis @jeremyfiel Thank you for the reviews! I'll definitely modify the examples as I do see how they can seem repetitive and basic. Will research more on how I can improve them with your suggestions and then update the PR soon!

pages/learn/json-schema-examples.md
spacether: How about using the publishedDate form instead? Having a space in a key is not something that I have seen widely used in json in my experience.

gregsdennis: Yes, that would fall under

> Otherwise, just use the actual property names.

aku1310:
I apologize for the delay in updating this PR 😅 

I have added some more properties to the examples such as `$dynamicRef`, `$anchor`, etc., and also removed some redundant examples which were showcasing something new.

Please let me know what you think about this.

components/Sidebar.tsx
```diff
@@ -39,7 +39,7 @@ const getDocsPath = [
 const getStartedPath = [
   '/learn/json-schema-examples',
   '/learn/file-system',
-  '/learn/miscellaneous-examples',
+  '/learn/generic-examples',
```
gregsdennis, line 42: I like miscellaneous here.  "Generic" already has specific meaning in programming, and we don't want to imply that these examples are that.

pages/learn/generic-examples.md
```diff
@@ -3,10 +3,9 @@ section: docs
 title: Miscellaneous Examples
 ---
 
-In this page, you will find examples illustrating different uses cases to help you get the most out of your JSON Schemas, including:
+In this page, you will find generic examples illustrating different uses cases to help you get the most out of your JSON Schemas. Each example comes with accompanying JSON data and explanation.
```
gregsdennis, line 6:
```suggestion
In this page, you will find miscellaneous examples illustrating different uses cases to help you get the most out of your JSON Schemas. Each example comes with accompanying JSON data and explanation.
```

pages/learn/json-schema-examples.md
```diff
@@ -92,129 +177,403 @@ title: JSON Schema examples
...
+  "summary": "Conference Presentation",
+  "location": "Conference Center",
+  "recurrenceDule": "FREQ=DAILY;COUNT=5"
+}
+```
+
+
+## Device type
+
+This schema represents electronic devices with a `deviceType` property that determines the device's category, such as `Smartphone` or `Laptop`.It employs the `$dynamicRef` keyword to dynamically reference schemas based on the `deviceType` property. This flexible schema structure allows data to conform to the appropriate device schema based on the `deviceType` specified, making it easy to describe different electronic devices with their unique characteristics
```
gregsdennis, line 195:
```suggestion
This schema represents electronic devices with a `deviceType` property that determines the device's category, such as `Smartphone` or `Laptop`. It employs the `$dynamicRef` keyword to dynamically reference schemas based on the `deviceType` property. This flexible schema structure allows data to conform to the appropriate device schema based on the `deviceType` specified, making it easy to describe different electronic devices with their unique characteristics
```

pages/learn/json-schema-examples.md
```diff
@@ -92,129 +177,403 @@ title: JSON Schema examples
...
+  "startDate": "2023-08-25T10:00:00Z",
+  "endDate": "2023-08-25T12:00:00Z",
+  "summary": "Conference Presentation",
+  "location": "Conference Center",
+  "recurrenceDule": "FREQ=DAILY;COUNT=5"
+}
+```
+
+
+## Device type
```
gregsdennis, line 193: I'm not sure what you're trying to do with this example.

The use of `$dynamicRef` is wrong,

- In 2020-12, `$dynamicRef` needs a `$dynamicAnchor` defined in the same schema resource, otherwise it works like an ordinary `$ref`.
- The value of `$dynamicRef` should be an anchor fragment (e.g. `#foo`), hence needing the anchor defined.
- The location `#/$defs/DeviceTypes` doesn't exist.

And I'm not sure a dynamic reference is the best solution for what the description says.  I'd use

- `oneOf` with `deviceType` defined using a `const` with a different value in each of the subschemas
- `if`/`then`/`else`
- follow [this example](https://json-schema.org/blog/posts/modelling-inheritance) to model inheritance (complex/advanced)

pages/learn/json-schema-examples.md
```diff
@@ -92,129 +177,403 @@ title: JSON Schema examples
...
+      }
+    },
+    "order": {
+      "$anchor": "OrderSchema",
+      "type": "object",
+      "properties": {
+        "orderId": { "type": "string" },
+        "items": {
+          "type": "array",
+          "items": { "$ref": "#/$defs/ProductSchema" }
```
gregsdennis, line 287: If you're going to go to the trouble of defining an anchor, you should use it.

```suggestion
          "items": { "$ref": "#ProductSchema" }
```

pages/learn/json-schema-examples.md
```diff
@@ -92,129 +177,403 @@ title: JSON Schema examples
...
+    "product": {
+      "$anchor": "ProductSchema",
+      "type": "object",
+      "properties": {
+        "name": { "type": "string" },
+        "price": { "type": "number", "minimum": 0 }
+      }
+    },
+    "order": {
+      "$anchor": "OrderSchema",
```
gregsdennis, line 281: This anchor isn't strictly necessary since it's not used anywhere, but leaving it isn't hurting anything.  (It might be used externally.)

pages/learn/json-schema-examples.md
```diff
@@ -92,129 +177,403 @@ title: JSON Schema examples
...
+      "type": "number",
+      "minimum": 0
     },
-    "bday": {
+    "applicationDeadline": {
+      "type": "string",
+      "format": "date"
+    }
+  },
+  "$anchor": "JobPosting"
```
gregsdennis, line 454: Having an anchor at the root isn't especially helpful.  You can do it, but why when

- internally you can just use `$ref: #` instead of `$ref: #JobPosting`
- externally, you can just `$ref: https://example.com/job-posting.schema.json` instead of `$ref: https://example.com/job-posting.schema.json#JobPosting`

pages/learn/json-schema-examples.md
```diff
@@ -92,129 +177,403 @@ title: JSON Schema examples
...
+    "cast": {
+      "type": "array",
+      "items": {
+        "type": "string"
+      },
+      "additionalItems": false
     }
-  }
+  },
+  "$anchor": "MovieSchema"
```
gregsdennis, line 510: Same comment as for `#JobPosting`

pages/learn/json-schema-examples.md
```diff
@@ -92,129 +177,403 @@ title: JSON Schema examples
...
-    "honorificSuffix": {
+    "medications": {
       "type": "array",
       "items": {
         "type": "string"
       }
     },
-    "nickname": {
+    "emergencyContact": {
+      "$ref": "https://example.com/user-profile.schema.json"
```
gregsdennis, line 394: I like the idea of interlinking the examples 👍

gregsdennis (changes requested):
It looks like there are some misunderstandings around `$dynamic*` and `$anchor`.  It'd be good to review how those work.  Unfortunately we don't really have good docs for `$dynamic*` just yet (that I'm aware of), but here's [`$anchor`](https://json-schema.org/understanding-json-schema/structuring.html#anchor).

aku1310:
> It looks like there are some misunderstandings around `$dynamic*` and `$anchor`. It'd be good to review how those work. Unfortunately we don't really have good docs for `$dynamic*` just yet (that I'm aware of) so you'll likely just need to try to read the spec (we can answer questions, too), but here's [`$anchor`](https://json-schema.org/understanding-json-schema/structuring.html#anchor).

Thank you! I'll go over this by the weekend, and update the PR soon.

components/Sidebar.tsx
aku1310: I had actually updated the term after discussing with Benjamin on the slack thread ([here](https://json-schema.slack.com/archives/CB3SS5W5B/p1693215510648819?thread_ts=1688030108.211109&cid=CB3SS5W5B)), since miscellaneous and other seemed a bit similar as titles.😅 
I can change it back to Miscellaneous if that seems better, please let me know how to proceed with this.

benjagm: > I can change it back to Miscellaneous

My bad @aku1310. @gregsdennis criteria has more weight here. Please let's rename it back to miscellaneous.

aku1310:
> Hey @aku1310, firsly, thanks again for taking the time and making the effort to try and improve our learning resources. Did you still want to continue to improve your PR here? If you're not wanting to spend much more time on it, maybe you could split some parts of it into a smaller PR, and then continue with other smaller PRs as time allows. Of course, please feel freeto ask further questions if you have any.

Hi, I apologise for the delay in this PR, I have not been able to properly work on this for a while 😅 Only the use cases in the other examples is left for now. Would it be alright to continue in this PR only? Or I can split it up if that is preferable. I aim to get this update the PR by this weekend as I have gotten free from other work now as well 😅

benjagm:
> Would it be alright to continue in this PR only?

That is totally fine Akanksha! We are very close to get this done. If you need help or feedback let me know and I'll be happy to help.

aku1310:
> That is totally fine Akanksha! We are very close to get this done. If you need help or feedback let me know and I'll be happy to help.

Will surely ask, thank you!!

aku1310:
> It looks like there are some misunderstandings around `$dynamic*` and `$anchor`. It'd be good to review how those work. Unfortunately we don't really have good docs for `$dynamic*` just yet (that I'm aware of) so you'll likely just need to try to read the spec (we can answer questions, too), but here's [`$anchor`](https://json-schema.org/understanding-json-schema/structuring.html#anchor).

Hi, as per the review, I have removed  `anchor` from `JobPosting` and `MovieSchema` as it was not helpful there, and also updated the `DeviceType` example using `oneOf` instead of `dynamicRef since `oneOf` does seems like a much better choice for that example. 

Please let me know what you think about this.

pages/learn/json-schema-examples.md
```diff
@@ -92,129 +177,414 @@ title: JSON Schema examples
...
+  },
+  "required": ["brand", "model", "processor", "ramSize"]
+}
+```
+
+**Data**
+
+```json
+{
+  "deviceType": "Smartphone",
```
gregsdennis, line 268: The casing here needs to match what the schema expects.  But the example looks a lot better now.

gregsdennis (approved):
Overall much better.  Left one more comment; just a minor thing.

pages/learn/json-schema-examples.md
aku1310: Oh yes, I'll just update that 😅

benjagm (approved):
Excellent job @aku1310!

Relequestual (approved):
Thanks for all your work on this. Very much appreciated!

Issue #87, opened 2023-10-03
UnderstandingJSONSchema.pdf is not longer regenerated


benjagm:
After unifying Understanding JSON Schema in the website we are no longer regenerating UnderstandingJSONSchema.pdf. However, this file has 700 weekly downloads, therefore we need to explore some alternatives:
- Explore ways to regenerate the file.
- Upload a pdf suggesting the people to visit the site or download the markdown files.

benjagm:
Resolved #122

PR #93, opened 2023-10-03
Add an initial version of a page on common interfaces.

Julian:
This page doesn't yet compile even, failing with some obscure Javascript error in a headline.

It was however ported from a reST page, so I'm still figuring out what flavor of markdown is needed to work with the site.

Marking it as [WIP] at least until I figure out what needs tweaking -- which certainly seems likely to be at least the `topic` and `glossary` bits, as Markdown doesn't really have equivalents of these, so I assume some Javascripting is necessary to implement something similar to the Bowtie page.

But the content is ready for immediate review.

### DIFF
--- a/pages/implementers/interfaces.md
+++ b/pages/implementers/interfaces.md
@@ -25,7 +25,7 @@ In addition to placeholder types like `String`, `Number`, `Boolean`, `Mapping`, 
 `Schema`
 
 > The type of JSON Schemas, which may also differ across dialects of JSON Schema.
-> For common or modern versions of JSON Schema this type is essentially `Map<String, Any> | Boolean`.
+> For common or modern versions of JSON Schema this type must essentially be able to represent both arbitrary JSON objects as well as booleans.
 
 `Instance`
 
--- a/pages/implementers/interfaces.md
+++ b/pages/implementers/interfaces.md
@@ -5,11 +5,11 @@ section: implementers
 
 JSON Schema is extremely widely used and nearly equally widely implemented.
 There are implementations of JSON Schema validation for [over 20 languages](https://json-schema.org/implementations.html) or environments.
-This prevalence is fantastic for user choice \-- as someone wishing to use JSON Schema you can be almost certain you\'ll find an implementation suitable for your target environment.
+This prevalence is fantastic for user choice -- as someone wishing to use JSON Schema you can be almost certain you'll find an implementation suitable for your target environment.
 
 But when it comes to community support, it can be challenging to know how to perform various JSON Schema operations in a particular library or implementation, as each may have slightly differing (or more than slightly differing) APIs.
 This can become particularly challenging when considering behavior not necessarily prescribed or required by the specification itself, but which is nonetheless common, useful, or even required behavior *in practice* in order to use JSON Schema.
-This page serves to document a number of these *common operations*, documenting the interfaces offered by many such implementations and inspired in part by the interfaces required for Bowtie\'s own operation.
+This page serves to document a number of these *common operations*, documenting the interfaces offered by many implementations across the existing JSON Schema ecosystem.
 
 For each, we first name and describe the operation using terminology from the specification where available, or otherwise using terminology that is aimed to be succinct but precise.
 The intention of this page is partially to be a reference of important JSON Schema operations as well as a way of tailoring help for a particular language and implementation.
@@ -30,7 +30,7 @@ In addition to placeholder types like `String`, `Number`, `Boolean`, `Mapping`, 
 `Instance`
 
 > The type of JSON instances.
-> This type should essentially be `Any` or a type capable of representing all JSON values, given JSON Schema\'s applicability to any representable JSON.
+> This type should essentially be `Any` or a type capable of representing all JSON values, given JSON Schema's applicability to any representable JSON.
 
 `Dialect`
 
@@ -39,13 +39,13 @@ In addition to placeholder types like `String`, `Number`, `Boolean`, `Mapping`, 
 
 `EvaluationContext`
 
-:   The type of \"fully ready\" schemas and instances *along* with any additional implementation-specific customizable behavior.
+:   The type of "fully ready" schemas and instances *along* with any additional implementation-specific customizable behavior.
     At minimum, this type is either `Schema → Instance` or `Schema × Instance` (depending on whether the implementation takes both schema and instance together or compiles schemas and then produces a separate function taking the instance to validate).
     It is highly likely to be richer in at least some of the following ways:
 
     > -   Given likely support for some form of schema registry in order to support referencing external schemas, this type will likely include a registry of some sort, i.e. `Mapping<URI, Schema> → Schema → ...`
     > -   If an implementation supports customizing which JSON types match to which language types (such as [discussed below](#type-customization)) then this type likely includes some representation of this mapping, i.e. `Mapping<String, Callable[...]> → Schema → ...` encapsulating what each type is mapped to during this evaluation.
-    > -   Similarly, if there is a specific API for [format assertion enablement](#format-assertion-enablement), some representation of the `format` keyword\'s behavior is present in the context
+    > -   Similarly, if there is a specific API for [format assertion enablement](#format-assertion-enablement), some representation of the `format` keyword's behavior is present in the context
     > -   If the implementation supports the [creation or customization of dialects](#dialect-creation), and especially if schemas can contain subschemas across different dialects, then the context will contain some representation of dialects, e.g. `Dialect → Schema → ...`
 
 `Result`
@@ -66,7 +66,7 @@ In addition to placeholder types like `String`, `Number`, `Boolean`, `Mapping`, 
 
 Because this page deals with concerns which cross language barriers, and different programming languages have different capabilities particularly around how they represent out of band errors (via exceptions, option types, wrapper return types, sentinel values or some other mechanism), this page also introduces[^1] specific notation for representing the types of functions which include explicit representation for their out-of-band errors.
 Interpreting these types will depend on the programming language.
-It\'s easier to explain the above with an example:
+It's easier to explain the above with an example:
 
 [^1]: The authors of this page are not aware of a suitable existing and/or pervasive notation which serves this purpose already.
       If you are aware of one, a pull request or pointer would be welcome.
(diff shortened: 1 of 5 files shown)


Julian:
The location of this page is also a bit awkward seeming for at least 2 reasons:

* one is #87 which basically seems to me to belong above this link but has a large number of links it's adding to the sidebar
* two is that this page is a reference but one directed at implementers, not end users, which no other page there really is meant for. ISTM an "Implementers" section is likely a decent idea, though maybe that not until we have *two* such pages rather than just one, so I've left it here for now. Obviously ideas on this or other things welcome.

benjagm:
Hi @Julian. I updated the PR que the logic to make the page compile and also I created a new section in the sidebar to better accommodate docs for implementers. I am working on creating a new icon for that section.

The page is now available here: https://interfaces.website-2v2.pages.dev/implementers/interfaces

benjagm (approved):

benjagm:
In the last version of code I am moving the link to this page back inside the Reference section, at the bottom in a sub-section called "For implementers" (see image)
Screenshot 2023-10-05 at 11 44 17

gregsdennis:
The preview is showing multiple instances of colons.  I'm not sure what these are for, but it looks like formatting that's not rendering properly.

image

Julian:
It's because there's no built-in markdown glossary "widget" so yeah it's formatting that's not rendering properly but that I don't yet know what to replace with. I guess probably I'll go with a table (that's what I meant by "which certainly seems likely to be at least the topic and glossary bits, as Markdown doesn't really have equivalents of these, so I assume some Javascripting is necessary to implement something similar to the Bowtie page." in the initial description)

pages/implementers/interfaces.md
```diff
@@ -0,0 +1,245 @@
...
+
+> The type of JSON instances.
+> This type should essentially be `Any` or a type capable of representing all JSON values, given JSON Schema\'s applicability to any representable JSON.
+
+`Dialect`
+
+> The type of JSON dialects or dialect identifiers.
+> This type may simply be `String` if the dialect is represented within the implementation by its URI or by some short name.
+
+`EvaluationContext`
```
gregsdennis, line 40: The description sounds like options/configuration to me.

For me, the "context" effectively manages a single evaluation.  It carries all of the data that a schema needs throughout the process and then gets thrown away.

Options include all of the things below (whether `format` validates, custom registries, etc).  In my implementation, the context _carries_ a copy of the options, but they are separate concepts.  My implementation only exposes the context 
to support users creating extension keywords.

benjagm:
> it looks like formatting that's not rendering properly.

Is `:::` Markdown? I am not sure about it.

gregsdennis:
> Is `:::` Markdown? I am not sure about it.

Yeah, I wonder if you're using some other markdown flavor.  `.rst`, maybe?

Just quickly googling...  yeah

> RST, Restructured Text, is a file format created by the Python community to write documentation. It is part of Docutils. RST is a markup language like HTML but is much more lightweight and easier to read.

You got python'd, @Julian. 😆

gregsdennis (commented):
Mostly content looks good.  Some formatting issues (I mentioned in another comment).  And some comments below.

Julian:
> Yeah, I wonder if you're using some other markdown flavor. .rst, maybe?

(reST is not markdown, but yes this page comes from an original written in reST :)

Thanks for the comments will have a look, and also obviously update when I figure out what to change the markup to.

pages/implementers/interfaces.md
```diff
@@ -0,0 +1,245 @@
+---
+title: Common Interfaces across JSON Schema Implementations
+section: implementers
+---
+
+JSON Schema is extremely widely used and nearly equally widely implemented.
+There are implementations of JSON Schema validation for [over 20 languages](https://json-schema.org/implementations.html) or environments.
+This prevalence is fantastic for user choice \-- as someone wishing to use JSON Schema you can be almost certain you\'ll find an implementation suitable for your target environment.
```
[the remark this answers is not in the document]
Julian: (It was inserted by pandoc during the conversion -- I'll remove it.)

pages/implementers/interfaces.md
```diff
@@ -0,0 +1,245 @@
...
+Contributions and corrections from implementers are particularly welcome, as are documentation links!
+
+For purposes of making some sections of this page more precise, we assume the existence of a set of *abstract types* all of which may be referenced below, and which will have specific concrete types within a given programming language or implementation.
+In addition to placeholder types like `String`, `Number`, `Boolean`, `Mapping`, `Callable` and the like, the JSON Schema-related types include:
+
+::: glossary
+`Schema`
+
+> The type of JSON Schemas, which may also differ across dialects of JSON Schema.
+> For common or modern versions of JSON Schema this type is essentially `Map<String, Any> | Boolean`.
```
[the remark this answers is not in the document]
Julian: A modern JSON Schema dialect allows you to define keywords whose values are any JSON value does it not?

Is "some type contained within `Map<String, Any> | Boolean` more agreeable?

pages/implementers/interfaces.md
Julian: Is there value in separating the two concepts? It seems simpler to flatten them and consider all of what you're calling options to be part of one big context which is indeed per-evaluation, otherwise the distinction needs to be explained at even more length -- whether you can do it on each evaluation or just once ahead of time is simply a programmer convenience is it not?

Julian: To be clear -- this is trying to be what you call "options", and not stateful for the evaluation -- i.e. it is not the place annotations are being built up inside.

(What word used seems malleable)

But I haven't here represented what you call "context" in your implementation at all I think, so I guess I'm asking whether you see value in adding a new concept here.

pages/implementers/interfaces.md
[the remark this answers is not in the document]
gregsdennis: > A modern JSON Schema dialect allows you to define keywords whose values are any JSON value does it not?

Yes, but that's not my confusion.  (C# doesn't have union types, so I'll drop the boolean for now.)

`Map<String, Any>` in C# is equivalent to `Dictionary<string, object>` (or if you want, `Dictionary<string, JsonNode>`), but that type is not at all convertible to my defined `JsonSchema` type, which is my model.  There are similarities, but they're quite different.

(That said, I do have an implicit cast defined that allows booleans to be convertible to some predefined global "true" and "false" schema constants.)

I think I'm just having a hard time abstracting.  Maybe it's okay as is, but the description doesn't really equate to what's going on for me.

pages/implementers/interfaces.md
gregsdennis: I don't necessarily see value in including what I call a context.  That's likely an implementation detail and (aside from the advanced use case of custom keywords) is completely hidden from the user.

However, I think "options" or "configuration" are likely better words for this idea.

gregsdennis:
@Julian another option is to support `.rst` on the website (and change your file extension).  Do we have an idea of how hard that is?

cc: @benjagm

benjagm:
> another option is to support `.rst`

Why not using a command line tool to convert your rst to markdown before adding it to the website? e.g. https://pandoc.org/

gregsdennis:
@benjagm I think it's already been through pandoc once.

https://github.com/json-schema-org/website/pull/93#discussion_r1348003585

benjagm:
@Julian can we merge this PR?

Julian:
It's still waiting for 2 fixes (from me) now that I'm back at work, and then for a review, so not quite yet.

pages/implementers/interfaces.md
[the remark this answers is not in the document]
Julian: I've weakened this a bit, see if what's there now matches what you're thinking.

Julian:
OK this is now likely ready for re-review, I've neutered some of the markup so it should render now.

pages/implementers/interfaces.md
```diff
@@ -0,0 +1,203 @@
...
+
+#### `URI`
+
+The type of RFC 3986 compliant URIs.
+This type should not generally be the same as `String`, as URIs have multiple possible string representations and require normalization to
+have correct semantics.
+
+---
+
+Because this page deals with concerns which cross language barriers, and different programming languages have different capabilities particularly around how they represent out of band errors (via exceptions, option types, wrapper return types, sentinel values or some other mechanism), this page also introduces[^1] specific notation for representing the types of functions which include explicit representation for their out-of-band errors.
```
gregsdennis, line 67: I wonder if this section might be better above the definitions.  The definitions (specifically `EvaluationOptions`) use your function syntax (which this section explains), but this section doesn't use the definitions at all.

gregsdennis (approved):
This reads a lot better, thanks.  Just a minor note.  Take or leave.

pages/implementers/interfaces.md
```diff
@@ -0,0 +1,203 @@
...
+
+## Version Detection
+
+**Type**: `Schema` → `Dialect`
+
+An API which identifies which dialect a given schema is written for, returning either the dialect itself or otherwise a dialect-specific validation function.
+
+## Type Customization
+
+An API which allows (re-)configuring which language-level types correspond to which JSON Schema types, and additionally potentially allowing for the definition of new types.
```
Relequestual, line 151: By this do you mean, like having a consturctor for the library so it can be used in multiple configurations simultaneously? Or, things can be re-configured? Or both?

Relequestual (approved):
One question, but otherwise this looks great. Thanks.

pages/implementers/interfaces.md
Julian: Whether it's per evaluation or once globally I suppose I didn't specify here (nor for other sections like the schema registry -- that too may either be something global or per-evaluation seemingly).

But the point is/was, if your programming language has multiple container types (e.g. a dynamic array, a linked list, and an immutable list, say) how do you configure which of those types map to the JSON `array` type (and therefore to JSON Schema's `items` keyword, say.)

pages/implementers/interfaces.md
Julian: The "new" thing in this section is meant to be the exception syntax (`<!> Foo`) not the function syntax itself -- which the definitions above don't use yet. But I'll re-read to see if this can go above.

pages/implementers/interfaces.md
Relequestual: > Whether it's per evaluation or once globally I suppose I didn't specify here (nor for other sections like the schema registry -- that too may either be something global or per-evaluation seemingly).

Ah, OK. I don't know that this element specifically matters in relation to JSON Schema. Although, I have seen problems where a library is unable to run multiple differnt dialects within the same file/scope. I realise this wasn't what you were getting at. I think there's globaly, per-evaluation, and per instance of a library (inbetween the two), which could apply to all configuration options and/or a schema registry. I could see the value in having both a global and instance based schema registry.

> But the point is/was, if your programming language has multiple container types (e.g. a dynamic array, a linked list, and an immutable list, say) how do you configure which of those types map to the JSON `array` type (and therefore to JSON Schema's `items` keyword, say.)

Oh, I'd never even considered that. My experience with typed langauges is limited, but is that sort of dynamic deserialization something that's common? (If I'm understanding this correctly, that's what you're suggesting?). I Imagin such would be quite tricky?

Julian: (Agreed on whether you have global state or not not being very interesting, that's why I didn't really mention it.)

---
> >  My experience with typed langauges is limited, but is that sort of dynamic deserialization something that's common?

My point of reference for this is my own implementation, where it looks like:

```
from jsonschema import validators
class MyInteger:
    ...
type_checker = Draft202012Validator.TYPE_CHECKER.redefine(
  "number",
  lambda _, value: Draft202012Validator.TYPE_CHECKER.is_type(value, "number") or isinstance(value, MyInteger),
)

CustomValidator = validators.extend(
    Draft202012Validator,
    type_checker=type_checker,
)
validator = CustomValidator(schema={"type" : "number"})
# instances of `MyInteger` are going to be treated as numbers
```

I don't know what kinds of languages are the ones where this would or wouldn't happen actually -or how common it is in implementations -- I'd personally have thought it'd be found in some as certainly in Python it's very long practice that JSON users want to extend it with new types (e.g. "how do I write a JSON array in JSON but then I want that to deserialize into a set in my programming language not an array"). Can look around though to see what other implementations have done in JSON Schema land.

Julian: > it sounds like you're saying there's an approach for saying "hey, you can cast this thing to a number" and processing native types as JSON for JSON Schema driven validation?

I think so :D -- I mean I'm somewhat confident I'm saying something concrete and that your rephrasing is the same thing, so I think so -- I think the vast *majority* of implementations do what you're calling "processing native types as JSON" if I parse those words correctly. A minority do "I'll parse a string into some types on your behalf" (which is represented on this page with the interface for "String based validation").

Relequestual: I think I'm on the same page now. I guess I just forget about these sorts of internals when I mostly work in javascript/node.js land.

PR #96, opened 2023-10-04
Fix links of understanding json schema links in landing page.

benjagm:
Related: https://json-schema.slack.com/archives/CB3SS5W5B/p1696433376276519?thread_ts=1694529805.273859&cid=CB3SS5W5B

### DIFF
--- a/pages/understanding-json-schema/_index.md
+++ b/pages/understanding-json-schema/_index.md
@@ -24,14 +24,14 @@ in the text.
 
 ## Where to begin?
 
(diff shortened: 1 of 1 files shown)


Julian:
I might recommend we look into a link checker rather than trying to fix these as one offs, it seems we've got a bunch of broken links and it's good to have CI check for them anyhow -- in Sphinx that's pretty trivial (one is built in) but perhaps we can find one in whatever framework we're using to transform markdown. If no one finds it first I can try to look.

PR #107, opened 2023-10-06
Add Plausible analytics

Relequestual:


Relequestual:
Need to add 404 tracking.
https://plausible.io/docs/error-pages-tracking-404

benjagm (approved):
Thanks a lot Ben!

Relequestual:
Turns out 404 tracking is tricker than it looks, or I'm just too new to using next.js/react and I don't understand something. Going to suggest NOT making that addition for now, so no longer blocks the PR.

benjagm (approved):
Thanks Ben!

Issue #152, opened 2023-10-13
No sitemap.xml


Relequestual:
Google is upset that many URLs that don't exist, result in a 404 status. (I'm not so upset by this!)

It does note that such upsets can be avoided if we provide a sitemap!

We might be able to use a module such as https://www.npmjs.com/package/next-sitemap to provide a sitemap with minimal effort, but I've not looked into this very deeply.

benjagm:
I am going to explore this other package: https://www.npmjs.com/package/globby

benjagm:
Finally I am using your package. It was easy to make it work in local but I found some challenges in Cloudflare. Finally it is working in this PR: https://github.com/json-schema-org/website/pull/155

benjagm:
Closed as completed.

PR #155, opened 2023-10-13
Add sitemap using next-sitemap

benjagm:
Related: #152


Relequestual:
Doesn't seem to show for the PR preview.
I tried two fixes based on https://community.cloudflare.com/t/sitemap-not-available-through-cloudflare-pages/384330/4
Neither have made the sitemap appear.

benjagm:
What fixed this was providing the build parameters in cloudflare because it is not using the next config of the repo.

Relequestual:
That's annoying. Can you add a new issue for someone to find out if you can make cloudflare use the config in the repo, and document in the readme whatever they find out please? =]

benjagm:
Hi Ben. What I meant was the Build setting. You can have build setting in your package.json but Cloudflare has its own build setting that you should configure, so what made this work in our case was the right code in next.config.js and next-sitemap.config.js and the right build commands on cloudflare. See image below:

Screenshot 2023-10-16 at 09 51 30

Relequestual:
Understood.
I think the build command should be in the package.json file though, then have the build command in Cloudflare call that.

We can then test making production builds locally, and have better control (in the repo) over the build command.

benjagm:
I have the same command in the package.json working locally too.

PR #173, opened 2023-10-23
Add an automatic commenter for new tooling PRs.

Julian:
This workflow will comment on PRs that add new tools / implementations to the website. It will do so only on PR open (i.e. once), not whenever the PR is modified.

The goal is to direct implementers (or users submitting on their behalf) to review some resources that might be applicable to them.

We can of course grow or shrink this list as needed, if we wish to point to additional resources.

Right now, the way it detects which PRs are tool additions is by considering any PR which touches:

    * the YAML files in the data directory
    * the pages/implementations/main.md page

to be possible candidates.

Additional tweaks could be made to add or remove these triggers if the commenter is too noisy, or not noisy enough.

For PRs which touch these pages, this looks like:

Screenshot 2023-10-23 at 1 01 49 PM

PRs which do not meet the above detection are ignored.

Some further context on the immediate motivator of this is [here](https://github.com/json-schema-org/community/issues/408#issuecomment-1775343633).

### DIFF
--- a/.github/workflows/new-implementation.yml
+++ b/.github/workflows/new-implementation.yml
@@ -1,0 +1,47 @@
+name: New Implementation Commenter
+on:
+  pull_request:
+    types:
+      - opened
+    branches:
+      - main
+    paths:
+      - data/*.yml
+      - pages/implementations/main.md
+
+permissions:
+  pull-requests: write
+
+jobs:
+  comment:
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/github-script@v6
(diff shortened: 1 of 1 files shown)


Julian:
Ooh, good idea, yes, adding.

gregsdennis (approved):

Julian:
@Relequestual will let you take a look in case you didn't see this one yet before merging in case you have any other things you can think of to add.

Relequestual (approved):
Thanks for this work!

PR #178, opened 2023-10-24
Remove rel=alternate links that 404

jdesrosiers:
Alternately (pun intended), I could create links to the actual XML documents. This is easier and I think it's sufficient.


benjagm (approved):
Thanks a lot for this PR Jason. Looks good to me.

Issue #170, opened 2023-10-23
Validation error in learning materials


Vaent:
Two examples of If-Then-Else logic are incorrectly showing as valid when they should be invalid. This has been discussed recently on the community Slack: https://json-schema.slack.com/archives/C5CF75URH/p1697791514589429

Affected page: https://github.com/json-schema-org/website/blob/main/pages/understanding-json-schema/reference/conditionals.md

Affected lines: 279, 287 (containing the "valid": true props in the incorrect examples).

```json
// props { "indent": true, "valid": true }
{
  "street_address": "24 Sussex Drive",
  "country": "Canada",
  "postal_code": "10000"
}
```
```json
// props { "indent": true, "valid": true }
{
  "street_address": "1600 Pennsylvania Avenue NW",
  "postal_code": "K1M 1M4"
}
```

PR #179, opened 2023-10-25
Fix error in if/then/else on UJS and cleanup

jdesrosiers:
Fixes https://github.com/json-schema-org/website/issues/170

In addition to fixing the error I did some editorial cleanup of the whole section.


benjagm (approved):
Looks good to me. Thanks for putting this together!

Relequestual (approved):

PR #182, opened 2023-10-26
Solves the issue 169 - Avoid the numbered to have all "1." in the getting started guide

### LINKED ISSUE #169: Numbered lists not rendering right on getting started page
https://json-schema.org/learn/getting-started-step-by-step

The numbered lists that are scattered throughout the page all have "1." instead of incrementing values.  Might be an issue with the markdown, but it might be a rendering thing.

benjagm:
Resolves #169 

Avoid the numbered to have all "1." in the getting started guide.

### DIFF
--- a/pages/learn/getting-started-step-by-step.md
+++ b/pages/learn/getting-started-step-by-step.md
@@ -99,7 +99,7 @@ Using the product catalog example, `productId` is a numeric value that uniquely 
 To add the `properties` object to the schema:
 
 1. Add the `properties` validation keyword to the end of the schema:
-
+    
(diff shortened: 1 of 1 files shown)


gregsdennis (approved):
Not sure why these changes fix it, but they seem to.

benjagm:
> Not sure why these changes fix it, but they seem to.

I compared this file with others where the ordered list was working properly and I found that increasing the indentation of each list item's content is what makes it work.  For this case the content was just a code snippet so adding an empty line with the increased indentation was enough to make it work.

PR #201, opened 2023-11-03
Update the links on a few pages to point to the new website repo.

Julian:
Fixes a few 404s, specifically on the Glossary page.

This does not update the pages/blog/posts/website-analytics-snapshot-2023.md 2023 blog post (which directly links to the json-schema-org.github.io source repository and references behavior that isn't the case for the new website). json-schema-org/community#514 is relevant here, and should likely consider what to do with this link and blog text as part of that PR.

This also does not touch files in the _include directory, which are submodules included from the spec repo and will have their own PR there.


benjagm (approved):
Thanks for adding this Julian.

PR #204, opened 2023-11-03
Migration of multiple PRs from Understanding JSON Schema repository

benjagm:
This PR contain the migration of multiple PRs from Understanding JSON Schema repository:

- https://github.com/json-schema-org/understanding-json-schema/pull/198
- https://github.com/json-schema-org/understanding-json-schema/pull/140
- https://github.com/json-schema-org/understanding-json-schema/pull/115


Relequestual (approved):
Thanks for taking care of these items.
Regarding the [Issue](https://github.com/json-schema-org/understanding-json-schema/pull/140) about adding an example of `$comment`.

The comment of the PR was also suggesting we might want to document that it does not work the same way as a traditional code commnet, as in it cannot override. For example, the meta-schema defines `$comment` in a `properties` object, and that is not a comment.

Happy for that to be a new Issue/PR

PR #206, opened 2023-11-05
chore: Update supported draft versions for Java Generator from code

CarstenWickner:
Support for Draft 2020-12 in the `victools`:`jsonschema-generator` library was introduced end of 2021; support for Draft 6 even before that.

Merely updating the reference here.


benjagm (approved):
Hi Carsten. Thanks a lot for updating this information!!
Have a great week!

PR #212, opened 2023-11-08
Upgrade dependencies

### LINKED ISSUE #113: Update node.js version
Currently the node.js version is pinned at `16.0.0`.
This means we are working on versions with known and fixed bugs.
(I haven't seen any evidence that anything is broken because of this however.)

Node.js 16 has reached end of life.
The active LTS is 18.
LTS will move to 20 on 2023-10-24.

My suggestion, if we can, is to move to version 20, although this might require upgrading some of the dependencies. If that proves to be challenging (like, it takes more than 30mins), then moving to version 18 might be easeir.

I'm happy to look into this, but also happy if someone beats me to it.

As an aside, and maybe this should be a separate issue, we are using Cloudflare pages builder version 1, and version 2 is available, with updated underlying OS to current TLS, and updated underlying tooling. We should additionally look to see if we can use that. This requires manually changing settings in the Cloudflare admin area.

jdesrosiers:
Fixes #113 

This brings all of the website's dependencies up-to-date :tada: including Node 20, Next.js 14, and React 18. There were a couple things that needed to be updated because of the upgrades, but the biggest result of the upgrades seems to be that it's doing a better job at warning about potential issues. Most of the changes in this PR are fixing those kinds of things.

When running the site in dev-mode, there's a warning about using `useLayoutEffect` on the server. Our site is client-side only, so that isn't a problem for us. I couldn't find a way to make that go away, so we'll see those messages in development, but it won't effect the site in production.


jdesrosiers:
I had to change a couple settings in Cloudflare Pages to get the preview to build.

I had to change the `NODE_VERSION` environment variable. I only changed the one for the build previews. We'll need to change the one for production _before_ merging this PR.

I also had to change the "Build command" to remove `next export`. That command no longer exists and is done automatically when `next build` is run. There isn't a separate "Build command" for preview and production, so I reverted my change after getting the preview to build. We'll need to remove `next export` _before_ merging this PR.

benjagm:
> remove `next export`

This was required to properly generate the sitemap files for SEO. This was the related discussion: https://github.com/json-schema-org/website/issues/152 

If we remove `next export` without providing an equivalent we'll stop having the sitemap properly generated and Google searches will start penalize.

benjagm:
> remove `next export`

I just added something to the PR and removed the `next export` command and everything seems to be working!

gregsdennis:
If this is working, can we go ahead and merge it?  I think other PRs will be waiting on this one.

benjagm:
> If this is working, can we go ahead and merge it? I think other PRs will be waiting on this one.

I agree that we need to merge this as soon as possible because this affects how the static files are generated and published in Cloudflare. To merge this we need to do some steps:

- Remove the `next export` arguments of the build settings in Cloudflare.
- Upgrade the node version of production to 20.XX.XX in Cloudflare.

Greg, the issues you faced with the blog PR was because I changed the build settings in Cloudflare to test this and I thought there shouldn't be side effects but it happened. I reverted the change and now you should be able to merge it.

gregsdennis:
Well mine still needs review.

Relequestual (approved):

benjagm (approved):

Relequestual:
Please feel free. I would do this if I didn't already need to finish something else today.

PR #221, opened 2023-11-12
Update dev.harrel:json-schema impl info (draft 2019-09 support)

harrel56:


Julian:
Blehhh it seems the commenter failed again on this PR!? I think that means something is wrong with permissions on *forked* repos?? But that makes no sense, because it should be trying to do things to this one... will have to take another look :(

Julian:
(Anyways the change lgtm so merging anyhow)

PR #222, opened 2023-11-13
Add a proposed mechanism for tracking known "quality" implementations.

Julian:
This PR adds a file for us to add data on "quality" implementations, for some defined characteristics of how we maight characterize such a thing. The file is nearly fully stubbed out -- so essentially the key content is the comment here, which we can use to farm this out hopefully.

Overall I tried to stress we're exposing data points here, not trying to (hopelessly) define a metric for what good and bad code/projects are.

There is obvious overlap here with our existing list of implementations -- note though that we have loose (or no) criteria for listing implementations on the page. Here instead we wish to specifically raise languages for which we know no good implementation to exist, despite the language being popular or widespread, if any.

What's here does *not* yet render this data anywhere, nor does it implement the dynamic API calling mentioned in the body (for retrieving statistics from any listed repository).

What it does do is identify a simple initial list of languages (from TIOBE and GitHub's own lists), and propose a format for tracking data points about implementations meeting the proposed criteria.

The initial goal is to get feedback on this template, and to solicit help in filling it in for these languages.

Refs: json-schema-org/community#418

### DIFF
--- a/data/validator-libraries-by-language.yml
+++ b/data/validator-libraries-by-language.yml
@@ -1,0 +1,90 @@
+# This file, unlike the adjacent validator list data files, collects metrics to try and assist in identifying "higher quality" implementations.
+#
+# Absence from this list is not meant to be a slight against anyone or anyone's work certainly.
+#
+# The goals are to constructively help users choose libraries for their desired environment, and to identify gaps in the ecosystem that might need help.
+# In that spirit, the collection of languages below is simply an initial list, based largely on:
+#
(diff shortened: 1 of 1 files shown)


benjagm:
Do we need this to appear in the website?

Relequestual:
> Do we need this to appear in the website?

Good question. It could just as easily live in the community repo. Let's discuss.

Julian:
Superceded (as per the OWCM call) via https://github.com/json-schema-org/community/pull/554

PR #223, opened 2023-11-14
Fix missing comma

Relequestual:


Julian (approved):
This LGTM too (though just in case, since I said it on some other PR, I'd opine that to me this kind of PR also isn't in need of review, if it were me, I'd have commit or merged it immediately as being trivially correct).

gregsdennis:
Agreed.

Issue #160, opened 2023-10-18
Re-build the JSON Schema Implementations page - (Design)


benjagm:
With the release of the new JSON Schema website, we still have plenty of range of improvement in the Implementations page. A new data-driven page, with filters and more call to actions to add implementations or try bowtie will improve the UX and the impact of this very important page.

Current page: https://json-schema.org/implementations

Inspirations:
- https://tools.openapis.org/
- LFX Landscapes: https://landscape.prestodb.io/card-mode

**Live Low-fi Prototype:** https://docs.google.com/presentation/d/1cCjQO0tdUqf74iEtYYv7E2dHxYyZbbzL_sEQ1_64W1Y/edit?usp=sharing

Slack thread: https://json-schema.slack.com/archives/CT8QRGTK5/p1697620659557929

benjagm:
Feedback provided by @Relequestual on Slack:

My preference here is as follows:
* For filters
  * When on Desktop, have them across the top, and sticky (stay when you scroll, and not on mobile)
  * For mobile, a filter button which opens an overlay with the filter options (like the main menu)

I don't want to take up more horizontal space. Some implementations have a lot of text, so the small width design might not work.
Should default to only showing draft-07 and above, but make it clear that a filter is enabled.
For categories, we should have some information that details what they mean. How and where we display that data can be discussed.
I know this is more of a technical detail, bot for many of the non-validation implementations, we do not know what drafts/versions are supported. We should have a filter for "unknown" or "undefined".
A link for adding an implementation is a good idea. Where we position that if we choose to put filters horizontally about the content, I'm not sure.
I think having a filter and maybe a column for Bowtie report ("Bowtie report (compliance)") would make sense. We would need to explain what that means.
For the "adding your tool", maybe we want "Something missing/wrong?" to include updates or changes and not just new tools.
Could do:
* We may want to add a filter for licence which is "All OSI approved licences"
* When selecting language/s, show number of results based on other filters (or grey out 0 result options)
* Reset all filters button
* Options in "drafts" to select all "modern" aka draft-07 and up, in one go

Aside: Do we want to say "drafts" or "versions"?
Under versions, we could say "draft-07" etc, but then "2019-09" and "2020-12" without "draft".
Great start! =]

big-andy-coates:
Hi @benjagm,

What's the plan for the Benchmarks section on the existing implementations page?

I ask as I totally missed there were benchmarks when I was researching which JVM based validator implementation to use (and ended up writing my own).  

[...]

PR #225, opened 2023-11-15
Add a link to new JVM validator benchmark

big-andy-coates:
Hey, it was suggested on [Slack](https://json-schema.slack.com/archives/C5CF75URH/p1697889828358549), that I raise a PR to add a link to the implementations page for a new performance benchmark & functionality comparison microsite we've created while trying to decide which JVM based validator implementation to use.

Site: https://www.creekservice.org/json-schema-validation-comparison/
Repo: https://github.com/creek-service/json-schema-validation-comparison

The site is designed to be auto-updating, via GitHub dependabot, as new versions of implementations are released. Contribution from implementation developers is encouraged.

### DIFF
--- a/pages/implementations/main.md
+++ b/pages/implementations/main.md
@@ -9,6 +9,9 @@ Benchmarks that compare at least two implementations supporting draft-06+ may be
 -   PHP
   -   [php-json-schema-bench](https://github.com/swaggest/php-json-schema-bench) - comparative benchmark for JSON-schema PHP validators using JSON-Schema Test Suite and z-schema/JSCK (MIT)
 
+-   JVM (Java, Kotlin, Scala)
+  -   [json-schema-validation-comparison](https://www.creekservice.org/json-schema-validation-comparison/) - an independent functional and performance comparison of JVM-based validator implementations using JSON-Schema Test Suite and more (MIT)
+
 ### API documentation
 
 -   JavaScript


big-andy-coates:
I also wanted to mention that I've previously never noticed the Benchmarks section when I've been researching which JVM based validator implementation to use.

I was therefore also wondering if there was some way we could add a link from the list of validators at the top of the page to the associated benchmark section. 

For example, add a link by the `Java`, `Kotlin` and `Scala`  heading to the `JVM` section under `Benchmarking`. The same could be done for the existing `PHP` and `JavaScript` sections. 

What do you think? Worth raising in an issue?

benjagm:
> I was therefore also wondering if there was some way we could add a link from the list of validators at the top of the page to the associated benchmark section.

Thanks a lot for this PR @big-andy-coates ! Your have done an amazing job with your tool and I am happy the community using Java will love it. We are working in a total rebuild of this page on this issue #160 . Do you mind sharing this feedback there? I am concerned that this won't be easy to implement in the new version that will be [totally data-driven with filters](https://docs.google.com/presentation/d/1cCjQO0tdUqf74iEtYYv7E2dHxYyZbbzL_sEQ1_64W1Y/edit?usp=sharing) so all implementation types has the same consideration. The good thing in the new approach is users using java will be able to filter by java and see all the options available for their needs, including benchmark.

benjagm (approved):
Thanks for adding your tool. I think is super cool what you have done!

PR #237, opened 2023-11-23
Update License to remove dual license clause

benjagm:
Update License to remove dual license clause.

Issue: n/a

ADR: Not required.

### DIFF
--- a/README.md
+++ b/README.md
@@ -134,5 +134,5 @@ Become a financial contributor and help us sustain our community. [[Contribute](
 This document has been inspired by [AsyncAPI website README.md](https://github.com/asyncapi/website/blob/master/README.md).
 
 ## License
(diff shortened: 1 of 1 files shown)


Relequestual:
After a chat with @benjagm, I understand this PR is correcting the readme, and the licence file only specifies AFL-3.0.

I'll move my comments out to a new issue.

Relequestual:
I'm noting here that the current licence file includes both BSD-3-Clause and AFL-3.0, so the readme is correct in saying that the repo is currently has multiple allowable licences.

benjagm:
> I'm noting here that the current licence file includes both BSD-3-Clause and AFL-3.0, so the readme is correct in saying that the repo is currently has multiple allowable licences.

As per this comment, we close this PR.

PR #247, opened 2023-11-25
Add `remark-lint-frontmatter-schema` to "Linters" implementations

JulianCataldo:
Hello :)

I didn't found a fitting pre-existing place for this Markdown oriented tool, so I've added an additional sub-section in "Utilities".

It works inside remark + remark-lint, so technically it's a **lint rule**, but also an IDE / pipeline runtime **validator** at the same time… 🤔
Tell me if it's looking good for you ;)

And, again, thanks for all the work you're putting up!


Relequestual:
Thanks for your contribuion @JulianCataldo! I think we might even use this. If not, we should be.

Maybe "linter plugin" would be the most appropriate category in this instance?

JulianCataldo:
Thx! Very much appreciated.

> Maybe "linter plugin" would be the most appropriate category in this instance?

"Linter plugin" does look good to me :) I've reflected this

@ota-meshi did an analogous plugin for ESLint that, I think, will fit perfectly too in this section, if he's interested to add it:
https://github.com/ota-meshi/eslint-plugin-json-schema-validator

benjagm (approved):
Thanks for improving the JSON Schema Ecosystem with this implementation!

Looks great to me!

Issue #253, opened 2023-11-30
Wrong link to Validation Specification for 2020-12


kool79:
1. Open https://json-schema.org/specification-links#published-drafts
2. Find **_Published Drafts > 2020-12 > Specification > Validation_** and click the link(s):

![scr_20231130T09_22_48](https://github.com/json-schema-org/website/assets/4625317/d56d0013-9a73-4193-8c3c-03e70464dfe4)

**_Expected:_** The page with documentation is opened:
 https://json-schema.org/draft/2020-12/draft-bhutton-json-schema-validation-01
 
 **_Actual:_** 404 error in browser console:
 
![scr_20231130T09_25_43](https://github.com/json-schema-org/website/assets/4625317/5f6f2c9e-f2fd-49e5-a878-9a59748b80b4)
and the home page is opened:
![scr_20231130T09_25_15](https://github.com/json-schema-org/website/assets/4625317/32820ba4-2574-4b71-9f55-86690f857cc8)

Notes: 
1. in browser console we have an url with .html.json at the end
2. note that address bar of homepage contains correct link (but with 'html'). When refresh the page, browser redirects to desired page (without html)

PR #254, opened 2023-11-30
Fix broken spec link

jdesrosiers:
Fixes https://github.com/json-schema-org/website/issues/253


jdesrosiers:
I consider it good project hygiene to update dependencies when ever I make changes, especially since dependency updates aren't automated in this repo. I can remove it if people aren't comfortable with that practice.

gregsdennis:
The policy I've learned professionally is that a change should just be the change and nothing else.  Isolating changes like this helps identify the source of problems when they do crop up.

If a change goes to production and causes an error, it can be hard to tell if it was the change or the updates that caused the problem.

I don't think it's a big deal for this PR, but just wanted to point it out.

gregsdennis (approved):

Relequestual:
I have to say from expereince working on more complex systems, I always put package updates in a seprate PR (or at least commit). Separate commits make it easier to do git-bisect to identify the causal commit of an issue down the road.

I don't think we need to change it for this PR, but I think it would be preferable moving forward.

PR #256, opened 2023-12-05
fix(numeric): fix wrong sample

vm-001:
**GitHub Issue:** # / NA

**Summary**:

Change `valid` to false, as the title said, "Numbers as strings are rejected"

Yes/No
[JUSTIFICATION]


jdesrosiers (approved):
Thanks for catching this error and for making the time to submit a fix!

PR #274, opened 2024-01-19
Blog: Understanding JSON Schema Lexical and Dynamic Scopes

jviotti:
A new Engineering post aiming to explain and illustrate JSON Schema Lexical and Dynamic scopes. A big reason for writing this is to polish my own understanding, so please put as much red ink as possible! 🙏🏻


jviotti:
@jdesrosiers @gregsdennis @Relequestual @Julian @benjagm I'd hugely appreciate your comments here!

pages/blog/posts/understanding-lexical-dynamic-scopes.md
```diff
@@ -0,0 +1,244 @@
...
+---
+
+Most of the keywords defined by the JSON Schema organization can be either
+evaluated on their own, or by only considering the values of their adjacent
+keywords. For example, the `type` keyword is independent of any other keywords,
+while the `minContains` and `maxContains` keywords depend on the `contains`
+keyword defined in the same schema object.
+
+However, there is a small set of keywords whose evaluation depend on the
+<i>scope</i> they are in. These keywords are `$ref`, `$dynamicRef`,
```
Julian, line 24: Any reason you're using tags instead of normal markdown syntax?

jviotti: They don't seem to work. `*` seems to remain as just literal `*` for some strange reason

pages/blog/posts/understanding-lexical-dynamic-scopes.md
```diff
@@ -0,0 +1,244 @@
...
+excerpt: "A deep dive on lexical and dynamic scopes, essential for understanding more advanced topics like dynamic referencing"
+---
+
+Most of the keywords defined by the JSON Schema organization can be either
+evaluated on their own, or by only considering the values of their adjacent
+keywords. For example, the `type` keyword is independent of any other keywords,
+while the `minContains` and `maxContains` keywords depend on the `contains`
+keyword defined in the same schema object.
+
+However, there is a small set of keywords whose evaluation depend on the
```
Julian, line 23:
```suggestion
However, there are a small set of keywords whose evaluation depend on the
```
I think this is common to both British and US English -- the rule (which I don't know by name by heart, I had to double check what it is, just to make it sound right) has to do with when you're using a collective noun (set) but really referring to each of its members individually, which is the case here.

pages/blog/posts/understanding-lexical-dynamic-scopes.md
```diff
@@ -0,0 +1,244 @@
...
+<i>scope</i> they are in. These keywords are `$ref`, `$dynamicRef`,
+`unevaluatedItems`, and `unevaluatedProperties`. Additionally, there is also a
```
Julian, line 25:
```suggestion
`unevaluatedItems`, and `unevaluatedProperties`. Additionally, there are also a
```
As above.

pages/blog/posts/understanding-lexical-dynamic-scopes.md
```diff
@@ -0,0 +1,244 @@
...
+set of keywords that <i>affect</i> the scope they are declared in. These keywords
+are `$id`, `$schema`, `$anchor`, and `$dynamicAnchor`.
+
+JSON Schema introduces two types of scopes: the <i>lexical scope</i> and the
```
Julian, line 29:
```suggestion
JSON Schema defines two types of scopes: the <i>lexical scope</i> and the
```

pages/blog/posts/understanding-lexical-dynamic-scopes.md
```diff
@@ -0,0 +1,244 @@
...
+<i>dynamic scope</i>. Understanding how these scopes work is essential for mastering
+some of the most advanced (and often confusing!) features of JSON Schema, such
+as dynamic referencing.
+
+Lexical Scope
+-------------
+
+JSON Schema is a circular data structure. What we mean by this is that as a
+schema author, you may use <i>applicator</i> keywords that take other schemas as
```
Julian, line 38: Maybe worth crosslinking to the glossary for some of what's here, though there are lots of terms still missing, I think `applicator` isn't yet there...

pages/blog/posts/understanding-lexical-dynamic-scopes.md
```diff
@@ -0,0 +1,244 @@
...
+arguments, effectively creating a tree of schemas. 
+
+Consider the following example. In the left, a JSON Schema consisting of 7
```
Julian, line 41:
```suggestion
Consider the following example. On the left, a JSON Schema consisting of 7
```

(n.b. -- is "left" right on mobile? Probably not and it's top/bottom?)

pages/blog/posts/understanding-lexical-dynamic-scopes.md
```diff
@@ -0,0 +1,244 @@
...
+subschemas (declared using the `properties` and `prefixItems` applicators). In
+the right, a tree representation of the schema, where each subschema
+corresponds to a node in the tree:
+
+![Thinking of a JSON Schema as a tree structure](/img/posts/2024/understanding-lexical-dynamic-scopes/tree.webp)
+
+Thinking of a schema as a tree structure greatly helps understanding lexical
```
Julian, line 48:
```suggestion
Thinking of a schema as a tree structure greatly helps in understanding lexical
```

pages/blog/posts/understanding-lexical-dynamic-scopes.md
```diff
@@ -0,0 +1,244 @@
...
+<i>When evaluation encounters a reference keyword, it abandons the lexical scope
+of the reference schema and enters the lexical scope of the destination
+schema</i>. In other words, the evaluation process needs to re-calculate the
+lexical scope of the schema it ends up at after following the reference.
+
+Consider the following example. In the left, the lexical scope of a subschema
+that defines a static reference. In the right, the lexical scope of the
+destination subschema after following such reference. Note that the destination
+is part of a different schema resource, under a different meta-schema. The
+change of lexical scope let's the destination schema be governed, as expected,
```
Julian, line 113:
```suggestion
change of lexical scope lets the destination schema be governed, as expected,
```

pages/blog/posts/understanding-lexical-dynamic-scopes.md
```diff
@@ -0,0 +1,244 @@
...
+resource. For example:
+
+![Following a remote reference to a nested schema resource](/img/posts/2024/understanding-lexical-dynamic-scopes/ref-anchor.webp)
+
+Dynamic Scope
+-------------
+
+To recap, the lexical scope of a schema consists of the schema itself and its
+corresponding parent schemas within the same schema resource. In comparison,
+<i>the dynamic scope of a schema consists of the subschemas that were evaluated
```
Julian, line 133: Minor, but might be more effective to emphasize *less* in this sentence, so the important bits stand out. E.g.:

```suggestion
the *dynamic* scope of a schema consists of the subschemas that were *evaluated so far*
```

pages/blog/posts/understanding-lexical-dynamic-scopes.md
```diff
@@ -0,0 +1,244 @@
...
+so far, including the schema itself, regardless of the schema resource they
+belong to</i>.
+
+Coming back to our analogy of a JSON Schema as a tree structure, the dynamic
+scope corresponds to the sequence of nodes that were visited by the evaluation
+process. In JSON Schema parlance, this is referred to as the <i>evaluation path</i>.
+
+Consider the following example. In the left, a JSON Schema and a matching JSON
```
Julian, line 141:
```suggestion
Consider the following example. On the left, a JSON Schema and a matching JSON
```

etc.

pages/blog/posts/understanding-lexical-dynamic-scopes.md
```diff
@@ -0,0 +1,244 @@
...
+instance with a single object property `bar`. In the right, the evaluation path
+on the JSON Schema tree representation for validating the JSON instance,
+following schema references accordingly:
+
+![A tree representation of the JSON Schema evaluation
+path](/img/posts/2024/understanding-lexical-dynamic-scopes/evaluation-path.webp)
+
+In comparison to the lexical scope, the dynamic scope of a schema cannot be
```
Julian, line 149:
```suggestion
In comparison to the lexical scope, the dynamic scope of a schema cannot always be statically determined
```

Julian (commented):
Nice looking post! I just did a first read through, mostly for flow, so I've only focused a tiny bit on the actual content. I'll probably re-read the rendered version for that.

Good topic though for more exposition!

pages/blog/posts/understanding-lexical-dynamic-scopes.md
Julian: Hah, fun. OK, weird. I don't know anything about the markdown compilation toolchain here, so guess if this works..

gregsdennis: I typically use _ for italics, but I don't see a reason why * shouldn't work as well.  Odd.

benjagm: This should be a problem with the website component in charge on converting Markdown to HTML. I'll file an issue for this.

benjagm: This PR fixes this bug: https://github.com/json-schema-org/website/pull/275

jviotti: @gregsdennis I tried _ as well and it didn't work either 😅

jviotti:
@benjagm Do we generate preview URLs for blog PRs? It'd be convenient for the team to see the rendered version.

jviotti:
> I'll probably re-read the rendered version for that.

@Julian In case it helps, here is a PDF of the original Markdown:

[Understanding JSON Schema Lexical and Dynamic Scopes.pdf](https://github.com/json-schema-org/website/files/13993762/Understanding.JSON.Schema.Lexical.and.Dynamic.Scopes.pdf)

benjagm:
> Do we generate preview URLs for blog PRs?

In theory Cloudflare should have been generated the preview link. Can you push a new change to see if it works?

pages/blog/posts/understanding-lexical-dynamic-scopes.md
```diff
@@ -0,0 +1,244 @@
...
+the evaluation of such reference (highlighted in blue):
+
+![Comparison of lexical scope and dynamic scope with a reference](/img/posts/2024/understanding-lexical-dynamic-scopes/references-dynamic.webp)
+
+### Beyond Schema Resource Boundaries
+
+In comparison to the lexical scope, the dynamic scope of a schema is not
+limited by its schema resource. Moreover, the dynamic scope is a singleton
+entity that remains for the entirety of the evaluation process. In other words,
+the dynamic scope <i>is</i> the complete evaluation path.
```
gregsdennis, line 204:
```suggestion
the dynamic scope the ordered set of resources through which the evaluation path passes.
```

pages/blog/posts/understanding-lexical-dynamic-scopes.md
```diff
@@ -0,0 +1,244 @@
...
+
+Consider a variation of the previous example, where this time the anchor is
+within a subschema of a different schema resource. The lexical scope of the
+destination schema drastically changes as it hits its schema resource barrier,
+where as the dynamic scope of the destination schema remains the same:
```
gregsdennis, line 209:
```suggestion
whereas the dynamic scope of the destination schema remains the same:
```

pages/blog/posts/understanding-lexical-dynamic-scopes.md
```diff
@@ -0,0 +1,244 @@
...
+the right, a tree representation of the schema, where each subschema
+corresponds to a node in the tree:
+
+![Thinking of a JSON Schema as a tree structure](/img/posts/2024/understanding-lexical-dynamic-scopes/tree.webp)
+
+Thinking of a schema as a tree structure greatly helps understanding lexical
+scopes. Under this analogy, the lexical scope of a schema consists of the
+<i>ancestors</i> of such node, including the node itself. In other words, the
+lexical scope of a schema consists of its parent schemas including the schema
+itself, but not its subschemas.
```
gregsdennis, line 52: I'm not sure that this is a sufficient definition for lexical scope.  In other words, why is lexical scope defined this way?  How does it benefit us to define it this way?

I think it may be helpful to consider a definition for what lexical scope may have been in draft 7, before we needed to distinguish it from dynamic scope.

pages/blog/posts/understanding-lexical-dynamic-scopes.md
```diff
@@ -0,0 +1,244 @@
...
+Summary
+-------
+
+Understanding how static and dynamic scopes work is essential for gaining a
+deeper understanding of JSON Schema. The most important points to remember are
+summarized in the following table:
+
+| Comparison Point                     | Lexical Scope | Dynamic Scope |
+|--------------------------------------|---------------|---------------|
+| **Definition**                       | Consists of the parent schemas including the schema itself, but not its subschemas | Consists of the subschemas that were evaluated so far, including the schema itself, but not its subschemas |
```
gregsdennis, line 225: Again, not sure how this definition of lexical scope is beneficial beyond distinguishing it from dynamic scope.

pages/blog/posts/understanding-lexical-dynamic-scopes.md
```diff
@@ -0,0 +1,244 @@
...
+to other schema resources, but across remote references as well.
+
+Summary
+-------
+
+Understanding how static and dynamic scopes work is essential for gaining a
+deeper understanding of JSON Schema. The most important points to remember are
+summarized in the following table:
+
+| Comparison Point                     | Lexical Scope | Dynamic Scope |
```
gregsdennis, line 223: It'll be interesting to see this table rendered on the web.  It's pretty wide.

pages/blog/posts/understanding-lexical-dynamic-scopes.md
```diff
@@ -0,0 +1,244 @@
...
+|--------------------------------------|---------------|---------------|
+| **Definition**                       | Consists of the parent schemas including the schema itself, but not its subschemas | Consists of the subschemas that were evaluated so far, including the schema itself, but not its subschemas |
+| **Determining the scope**            | Can be statically determined without taking instances into account | Cannot be always statically determined. It may vary depending on the instance |
+| **Following references**             | Consists in abandoning the lexical scope of the reference schema and entering the lexical scope of the destination schema | Consists in retaining the current dynamic scope and adding the destination schema to it |
```
gregsdennis, line 227:
```suggestion
| **Following references**             | Abandons the lexical scope of the reference schema in favor of that of the destination schema | Appends the destination resource to the current dynamic scope |
```

gregsdennis (changes requested):
Overall, really good.  I think we need to revisit the definition of lexical scope.  The current definition seems to only serve as a point of distinction from dynamic scope, but lexical scope existed in previous spec versions.  We should consider what it meant in those specs and why.

pages/blog/posts/understanding-lexical-dynamic-scopes.md
Julian: It doesn't sound quite right to me (in American English) as I say, I think this fits under cases where you're referring to each individual keyword, in which case a noun like `set` still gets plural in American English -- but it doesn't matter much, I'm sure leaving it as-is is fine too so feel free to ignore.

benjagm:
> Do we generate preview URLs for blog PRs?

I found this in Cloudflare support: [Link](https://community.cloudflare.com/t/pages-preview-deployments-not-triggered-by-prs-from-forks/251385)

It seems that the preview links does not work if the PRs comes from a fork.

jviotti:
Thanks for the great comments so far. Let me accept and resolve the easy ones (on grammar, etc) and then I'll focus more on the content ones.

pages/blog/posts/understanding-lexical-dynamic-scopes.md
jviotti: Makes sense!

pages/blog/posts/understanding-lexical-dynamic-scopes.md
jviotti: Makes sense!

pages/blog/posts/understanding-lexical-dynamic-scopes.md
jviotti: This is how it looks like. I think its OK, so I'll resolve this thread, but let me know if not!

Screenshot 2024-01-22 at 11 22 37

jviotti:
@benjagm There also seems to be some issues with the outline at the left. It is showing a strange "Image by" artifact?

Screenshot 2024-01-22 at 11 25 00

pages/blog/posts/understanding-lexical-dynamic-scopes.md
jviotti: Yeah, good point. Let me cross link to anything that's already defined in there. I agree it'd be great to expand the list, so I can also cross link more from learnjsonschema.com

pages/blog/posts/understanding-lexical-dynamic-scopes.md
```diff
@@ -0,0 +1,248 @@
...
+tags:
+  - Fundamentals
+  - Scopes
+type: Engineering
+cover: /img/posts/2024/understanding-lexical-dynamic-scopes/cover.webp
+authors:
+  - name: Juan Cruz Viotti
+    photo: /img/avatars/jviotti.webp
+    link: https://www.jviotti.com
+    byline: O'Reilly author, founder of Intelligence.AI and Sourcemeta
```
Relequestual, line 13: I assume, like others, I may highlight `Intelligence.AI` and then open it in a new tab. The resulting site is not the one I'm looking for.

The non-www version of that URL is some weird gambling site... Assuming you have full control over the domain, might want to fix that and setup a redirect!

pages/blog/posts/understanding-lexical-dynamic-scopes.md
```diff
@@ -0,0 +1,244 @@
...
+excerpt: "A deep dive on lexical and dynamic scopes, essential for understanding more advanced topics like dynamic referencing"
+---
+
+Most of the keywords defined by the JSON Schema organization can be either
+evaluated on their own, or by only considering the values of their adjacent
+keywords. For example, the `type` keyword is independent of any other keywords,
+while the `minContains` and `maxContains` keywords depend on the `contains`
+keyword defined in the same schema object.
```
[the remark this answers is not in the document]
jviotti: Great point. Thanks. Done

pages/blog/posts/understanding-lexical-dynamic-scopes.md
jviotti: Oh, what the hell. Yes, we own the domain. Not sure how this could be taking place there... I'll look into it cc @RonItelman

pages/blog/posts/understanding-lexical-dynamic-scopes.md
```diff
@@ -0,0 +1,244 @@
...
+
+Consider the following example. In the left, the lexical scope of a subschema
+that defines a static reference. In the right, the lexical scope of the
+destination subschema after following such reference. Note that the destination
+is part of a different schema resource, under a different meta-schema. The
+change of lexical scope let's the destination schema be governed, as expected,
+by its own meta-schema instead of by the meta-schema of the root schema
+resource:
+
+![Lexical scope before and after following a static reference](/img/posts/2024/understanding-lexical-dynamic-scopes/ref.webp)
```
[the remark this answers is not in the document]
jviotti: Agreed that it is discouraged. The intention was to explain how scoping would work under the scenario, as I think it highlights the point well.

Maybe enough to clarify that it is indeed discouraged? Or do we still want to completely remove the example? Thoughts from the rest?

pages/blog/posts/understanding-lexical-dynamic-scopes.md
jviotti: > I think it may be helpful to consider a definition for what lexical scope may have been in draft 7, before we needed to distinguish it from dynamic scope.

Can you link me to this definition? I'm not very familiar with Draft 7, and the specification doesn't seem to mention lexical scopes or even scope in it. Is there a specific section you have in mind?

pages/blog/posts/understanding-lexical-dynamic-scopes.md
jviotti: The point of the article was to exemplify lexical & dynamic scopes by themselves, not necessarily showcasing any specific use cases (i.e. we don't dive into dynamic referencing, which is the only use case of dynamic scopes at this point if I understand correctly).

Is there another kind of definition that you have in place? Might be related to your Draft 7 comment above, which I don't think I fully understand yet. Can you provide more info on this? What kind of definition would you expect?

benjagm:
> It is showing a strange "Image by from" artifact?

I found that is related to the last <i></i> you added at footnote. This will be fixed as soon as you fetch the last changes and replace <i></i> with the regular italic markdown. It worked for me.

pages/blog/posts/understanding-lexical-dynamic-scopes.md
```diff
@@ -0,0 +1,252 @@
...
+![Following a remote reference to a nested schema resource](/img/posts/2024/understanding-lexical-dynamic-scopes/ref-anchor.webp)
+
+Dynamic Scope
+-------------
+
+To recap, the lexical scope of a schema consists of the schema itself and its
+corresponding parent schemas within the same schema resource. In comparison,
+the *dynamic* scope of a schema consists of the subschemas that were *evaluated so far*, 
+including the schema itself, regardless of the schema resource they
+belong to.
```
jdesrosiers, line 143: Not quite. The lexical scope is also the set of subschemas that were evaluated, it's just limited to the current schema resource.

jviotti:
True. Done!

pages/blog/posts/understanding-lexical-dynamic-scopes.md
```diff
@@ -0,0 +1,252 @@
...
+reference involves retaining the current dynamic scope, and adding the
+destination schema to it</i>.
+
+Consider the following example of a schema that follows a reference to an
+anchor. On the left, the lexical scope of the destination schema (highlighted
+in red), consists of it parent schema (highlighted in green), and its
+grandparent schema (highlighted in blue). However, on the right, the dynamic
+scope of the destination schema (highlighted in red) consists of the schema
+that referenced the anchor (highlighted in green) and the schema that led to
+the evaluation of such reference (highlighted in blue):
```
jdesrosiers, line 203: This isn't correct. The schema highlighted in green is not part of the lexical scope. This example is one where the lexical and dynamic scopes should align.

jdesrosiers (changes requested):
This is a great start, but I think there's a slight misunderstanding of lexical scope. You describe dynamic scope perfectly, but then describe lexical scope as something slightly different. Lexical and dynamic scopes are determined the same way. The lexical scope is just a subset of the dynamic scope that includes only the locations within the current schema resource.

pages/blog/posts/understanding-lexical-dynamic-scopes.md
gregsdennis: I think that was kinda my point.  Lexical scope definitely existed before we added dynamic scope, but we didn't have a name for it until we needed to distinguish between the two ideas.

This may be something that we need more discussion around, but if others are happy with this for the purpose of this blog post, I'm okay with it.

Probably still something to consider adding to the spec and to the glossary.

gregsdennis: From @jdesrosiers's [comment](https://github.com/json-schema-org/website/pull/274#pullrequestreview-1837060119) below:

> The lexical scope is just a subset of the dynamic scope that includes only the locations within the current schema resource.

I was also thinking the lexical scope was just the local resource, but in reading the rest of the post, I realized that some portions depend on your definition.  Rather than propose this, I thought I'd just ask why you defined it that way.

pages/blog/posts/understanding-lexical-dynamic-scopes.md
```diff
@@ -0,0 +1,244 @@
...
+
+To recap, the lexical scope of a schema consists of the schema itself and its
+corresponding parent schemas within the same schema resource. In comparison,
+<i>the dynamic scope of a schema consists of the subschemas that were evaluated
+so far, including the schema itself, regardless of the schema resource they
+belong to</i>.
+
+Coming back to our analogy of a JSON Schema as a tree structure, the dynamic
+scope corresponds to the sequence of nodes that were visited by the evaluation
+process. In JSON Schema parlance, this is referred to as the <i>evaluation path</i>.
```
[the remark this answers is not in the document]
jdesrosiers: I don't think "resources" is correct when discussing the general concept of dynamic scope rather than the implementation of dynamic references where it's more practical to track resources.

The dynamic scope is the stack of subschema locations visited while evaluating the schema. Anchors/dynamicAnchor a weird class of keywords that apply whenever you enter a resource (also includes `$id` and `$schema`). So, no matter what subschema I enter the resource through, any anchors in that resource are going to be in scope. That's why it's only necessary to track resources to implement dynamic references even though the dynamic scope is actually the visited subschemas.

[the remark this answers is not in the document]
jdesrosiers: (I'll add that I hate that dynamic scope is defined this way. It's unhelpful and unnecessary. It only serves to make dynamic references harder to understand. It should be a resource-level concept IMO.)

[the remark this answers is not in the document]
jviotti: > The dynamic scope is the stack of subschema locations visited while evaluating the schema. 

So if I'm following correctly, is the definition I'm providing in the article technically correct (even if we want to explain it in some other way)? By what @jdesrosiers is saying, it is indeed the stack of visited subschemas, not the stack of schema resources

pages/blog/posts/understanding-lexical-dynamic-scopes.md
jviotti: > I thought I'd just ask why you defined it that way.

I'm just trying to write down what I understand so far, so it's good that we have these discussions. If there is a better way to explain the ideas, in a way that I might have missed, then that's great and we should do that.

> The lexical scope is just a subset of the dynamic scope that includes only the locations within the current schema resource.

That's a very interesting observation, but not sure I fully get it. Lexical & dynamic scopes will typically diverge when references are encountered. There would be schemas from the schema resource that are not in the dynamic scope but would be on the lexical scope, so lexical scope wouldn't always be a subset of the dynamic scope, correct?

pages/blog/posts/understanding-lexical-dynamic-scopes.md
```diff
@@ -0,0 +1,252 @@
...
+![Defining multiple anchors with the same name on different schema resources](/img/posts/2024/understanding-lexical-dynamic-scopes/static-anchors.webp)
+
+### Changing Scopes With References
+
+Now that we understand lexical scopes and their boundaries within schema
+resources, lets consider the effect that following references has on them.
+<i>When evaluation encounters a reference keyword, it abandons the lexical
+scope of the reference schema and enters the lexical scope of the destination
+schema</i>. In other words, the evaluation process needs to re-calculate the
+lexical scope of the schema it ends up at after following the reference.
```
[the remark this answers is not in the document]
jviotti: > The subschema you end up in is simply the beginning of the lexical scope for that document

This is the part I guess I'm confused about. I thought the lexical scope consists of the parents of the schema, and that's why i.e. if you jump to a subschema of a schema resource that defines its own `$schema` you inherit that new dialect (because `$schema` was on the lexical scope of the target).

jviotti:
> You describe dynamic scope perfectly, but then describe lexical scope as something slightly different. 

I think I'm a bit confused. The current state of the reviews are:

- @gregsdennis points out that I could explain the lexical scope better while the current dynamic scope explanation is incorrect, while
- @jdesrosiers points out that the lexical scope explanation is incorrect while the dynamic scope one seems to be correct

Maybe there is a deeper misunderstanding here? I'm all for fully re-writing the post if needed and revamp how both of them are explained (to polish my own understanding), but I'd like to avoid going on one direction if there is some internal misalignment on these definitions.

jviotti:
On lexical scopes, I think @jdesrosiers is touching on the root of my potential confusion here: https://github.com/json-schema-org/website/pull/274#discussion_r1462268795

pages/blog/posts/understanding-lexical-dynamic-scopes.md
gregsdennis: Maybe "the lexical scope is the current schema resource" is a better way to put it.  @jdesrosiers , thoughts?

pages/blog/posts/understanding-lexical-dynamic-scopes.md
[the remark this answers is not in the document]
jviotti: This is the part of the spec I'm referring to:

> Note that some keywords, such as "$schema", apply to the lexical scope of the entire schema resource, and therefore MUST only appear in a schema resource's root schema.
> https://json-schema.org/draft/2020-12/json-schema-core#section-7.1-4

That seems to imply that `$schema` applies to a lexical scope because its one of the structural parents of the schema, but maybe I'm getting it wrong

pages/blog/posts/understanding-lexical-dynamic-scopes.md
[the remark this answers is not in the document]
gregsdennis: > That's why it's only necessary to track resources to implement dynamic references even though the dynamic scope is actually the visited subschemas. @jdesrosiers 

What's the benefit of defining a dynamic resource in a way that doesn't help you resolve dynamic references?  Supporting dynamic references is _why_ we defined dynamic scope.

> So, no matter what subschema I enter the resource through, any anchors in that resource are going to be in scope.

If the anchors _anywhere in the resource_ are in scope, then that tells me that the entire resource must be part of the scope.  Thus dynamic scope is the ordered set of resources identified by the evaluation path.

pages/blog/posts/understanding-lexical-dynamic-scopes.md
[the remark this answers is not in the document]
jviotti: Also, on why I used the term "calculate" here:

> While following the reference keyword moves processing from one lexical scope into a different one
> https://json-schema.org/draft/2020-12/json-schema-core#section-7.1-6

The specification seems to say we "move" from one lexical scope to another one when following any reference (not just dynamic), and to me "move" means "recalculate what the other one is" instead of just "discard the one you have until that point"

There is also this part:

> A keyword on the far side of that reference that resolves information through the dynamic scope will consider the originating side of the reference to be their dynamic parent, rather than examining the local lexically enclosing parent.

Which contrasts how the dynamic scope is more about the evaluation path while the lexical scope is more about its "enclosing parent"

pages/blog/posts/understanding-lexical-dynamic-scopes.md
```diff
@@ -0,0 +1,252 @@
...
+`/properties/bar/prefixItems/0` (highlighted in red) consists of the set of its
+parent schemas. The subschemas that are not part of its lexical scope have been
+grayed out:
+
+![The lexical scope of a JSON Schema represented as a tree structure](/img/posts/2024/understanding-lexical-dynamic-scopes/ancestor.webp)
+
+In comparison to the dynamic scope, the lexical scope of any subschema can be
+statically determined without taking
+[instances](https://json-schema.org/learn/glossary#instance) into account, just
+as we did here.
```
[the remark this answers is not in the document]
jviotti: Hm, I guess I'm confused then. The spec seems to only mention the evaluation path + instance with respects to the dynamic scope, and never with respects to the lexical scope:

> Other keywords may take into account the dynamic scope that exists during the evaluation of a schema, typically together with an instance document.

which made me think the distinction only applied to the dynamic scope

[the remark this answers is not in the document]
jviotti: There is also this part of the spec:

> The lexical scope of a keyword is determined by the nested JSON data structure of objects and arrays.

In conjunction with:

> Other keywords may take into account the dynamic scope that exists during the evaluation of a schema, typically together with an instance document. The outermost dynamic scope is the schema object at which processing begins, even if it is not a schema resource root. The path from this root schema to any particular keyword (that includes any "$ref" and "$dynamicRef" keywords that may have been resolved) is considered the keyword's "validation path."

Which made me think the lexical scope is about the structure of the schema while the dynamic scope was about the validation/evaluation path

pages/blog/posts/understanding-lexical-dynamic-scopes.md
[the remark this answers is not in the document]
gregsdennis: It may help to understand that "lexically" just means "as written".  So here, "local lexically enclosing parent" doesn't imply that the parent is part of the scope, but rather that a dynamic ref doesn't consider the parent "as written" like a static ref might.

It is rather confusing to make that clarification if the lexical scope doesn't include the local parent, though.

pages/blog/posts/understanding-lexical-dynamic-scopes.md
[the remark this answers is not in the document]
jviotti: Also on this part:

> A keyword on the far side of that reference that resolves information through the dynamic scope will consider the originating side of the reference to be their dynamic parent, rather than examining the local lexically enclosing parent.

The spec seems to say that the origin of a reference (and note it never says whether across schema resources or not) seems to only be taken into account for the dynamic scope, which is the reason I didn't consider the lexical scope as the evaluation path too, as you are pointing out on this example.

Anyway, not trying to say you are wrong, but trying to echo out loud what parts of the spec are confusing me

pages/blog/posts/understanding-lexical-dynamic-scopes.md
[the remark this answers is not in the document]
jdesrosiers: @gregsdennis, I agree with everything you said. It doesn't make a lot of sense. You'll have to ask Henry why he defined it that way.

pages/blog/posts/understanding-lexical-dynamic-scopes.md
jdesrosiers: > There would be schemas from the schema resource that are not in the dynamic scope but would be on the lexical scope

This is not correct. The lexical and dynamic scope are constructed in exactly the same way. There's no way the lexical scope can differ from the dynamic scope other than the resource boundary restrictions.

gregsdennis:
If anything, this conversation is telling me that we need to add things like "scope" to the definitions section in the spec.

pages/blog/posts/understanding-lexical-dynamic-scopes.md
jdesrosiers: > Maybe "the lexical scope is the current schema resource" is a better way to put it. @jdesrosiers , thoughts?

It _should_ be that simple, but I don't think that's technically the way dynamic scope is defined. If I was writing an article about dynamic references, that's what I would say as a little white lie because that's a far more reasonable way to explain it in the context of dynamic references and dynamic references are the only use-case for dynamic scope. Since this article is about dynamic scope, not dynamic references, I probably wouldn't go with that description here.

pages/blog/posts/understanding-lexical-dynamic-scopes.md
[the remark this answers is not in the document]
jdesrosiers: I mentioned in another comment that some keywords have a quality of applying to the whole document. These include `$id`, `$schema`, `$anchor`, and `$dynamicAchor`. Whatever subschema you're in, those keywords are in scope. So, when you enter `https://my-dialect.com#/items`, the `$id` and `$schema` are part of the evaluation of that schema even tho they are in a different sub-schema.

It's not as simple as saying the parents schemas are in the same scope, because it's very possible that there are schemas in between the root with `$id` and `$schema` and the schema you're evaluating that are never part of the scope.

pages/blog/posts/understanding-lexical-dynamic-scopes.md
gregsdennis: But what's the point of defining scope outside of the context of references?  Scope exists solely to define how references are resolved.  It seems, then, that scope needs to be defined with that in mind.

If this were my task, I would define:

- Lexical scope is the resource that contains the subschema/keyword being evaluated.  (A fragment-only URI points to the same lexical scope, whereas a non-fragment-only URI indicates a new scope to resolve.)
- Dynamic scope is the ordered set of resources through which evaluation has passed.

These definitions allow you to properly resolve both kinds of references, and I think that is sufficient for the purpose of this post.

pages/blog/posts/understanding-lexical-dynamic-scopes.md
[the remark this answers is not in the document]
gregsdennis: @jdesrosiers I think you're mis-reading the schema Juan posted.  `https://my-dialect.com` is a new `$schema`, not the `$id` of the subschema.

pages/blog/posts/understanding-lexical-dynamic-scopes.md
jviotti: FWIW, @gregsdennis 's explanation above is also quite simple/elegant. Happy to go with it if we all agree

pages/blog/posts/understanding-lexical-dynamic-scopes.md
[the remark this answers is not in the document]
jviotti: > Whatever subschema you're in, those keywords are in scope

That sounds to me like yet a 3rd type of scoping. Risking making things more complicated, maybe we should give this a name?

jviotti:
> I'm thinking more and more that it probably doesn't make sense to try to describe lexical/dynamic scopes. 

IMHO, that feels like bypassing the problem. The specifications do define "scopes" as specific entities and readers attempting to understand them are likely bound to fall for the same confusions I fell into. I agree its hard to define them properly (as there also seems to be some ambiguity within us of what they should be), but I also feel we might get somewhere and accomplish both things: an actual concrete definition that is useful for thinking about references.

I think this comment gets quite close already: https://github.com/json-schema-org/website/pull/274#discussion_r1462462689

Relequestual (changes requested):
I've left a few comments. Overall this is great and is going to be extreamly useful in terms of clearly communicating how to understand scope to the community. Thank you so much for doing this work. (I know there's promoting the book work, but still, thanks!)

There's something I'm not sure is 100% accurate, which I don't think I covered in my comments, because it touches the whole understanding of scoping.

I need to find a test or do some manual testing myself, to validate what I have in mind.

Specifically, I think lexical scope includes ancestors and decendants within the same schema resource (but not crossing resource boundaries).

I think I may have spoken with Henry about this before. I haven't (yet) been able to find a test, and I really should have some schema to explain what I mean, or it may not make sense. But if I'm right, several parts of your explanation will need to be changed here, and maybe one or two of the images.

Again, thanks for this work! Super valuable!

jviotti:
> I think I may have spoken with Henry about this before. I haven't (yet) been able to find a test, and I really should have some schema to explain what I mean, or it may not make sense. But if I'm right, several parts of your explanation will need to be changed here, and maybe one or two of the images.

@Relequestual Awesome! Looking forward to that. Right now I want to collect all of this information from the team, and then I can plan how to revamp the article accordingly!

pages/blog/posts/understanding-lexical-dynamic-scopes.md
jviotti: @Relequestual Based on some of your comments, how do you feel with Greg's definition above ^?

jviotti: Maybe one potential minor comment from me:

> Dynamic scope is the ordered set of resources through which evaluation has passed.

If we are saying that the lexical scope if the entire schema resource, should we say that the dynamic scope is the ordered set of "subschemas" instead?

Then, questions around how `$schema` and `$anchor` come into play are resolved by saying that they are within the lexical scope of each of the subschemas in the dynamic scope. Seems to complement it well on my mind

gregsdennis: > within the lexical scope of each of the subschemas in the dynamic scope

Moving forward, we're removing `$dynamicRef`'s ability to act like a regular `$ref`, meaning that the two referencing mechanics will be completely disjoint.  I think it's easier to understand them this way, and defining one based on the other just confuses things.

Just my opinion, though.

pages/blog/posts/understanding-lexical-dynamic-scopes.md
```diff
@@ -0,0 +1,251 @@
...
+
+JSON Schema defines two types of scopes: the *lexical scope* and the *dynamic
+scope*. Understanding how these scopes work is essential for mastering some of
+the most advanced (and often confusing!) features of JSON Schema, such as
+dynamic referencing.
+
+Lexical Scope
+-------------
+
+JSON Schema is a circular data structure. What we mean by this is that as a
```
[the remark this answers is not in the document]
gregsdennis: Yeah, the spec actually instructs implementors to guard against circular references.  Good point.

pages/blog/posts/understanding-lexical-dynamic-scopes.md
jviotti: Ah, makes sense then. What I'll do is probably leave this PR as-is for a couple more days in case anybody else has any thoughts, then push for your definition above ^, and then address the content accordingly

jdesrosiers: @jviotti @gregsdennis @Relequestual There are so many threads going on, so I'm going to respond to the general themes here. If there's anything I don't cover that you want me to respond to, please ping me.

### Lexical scope and instances
I probably misunderstood Henry and the lexical scope is not related to the evaluation path. Even if I didn't misunderstand Henry, I don't think that interpretation makes sense and I'm more than happy with us all deciding to describe it in a different way. However, in that case, I don't see how the concept of lexical and dynamic scopes converging makes sense. These are two completely different mechanisms doing different things, so how can they ever be the same?

### Dynamic scope: subschemas or resources
I feel that Henry was quite clear that the dynamic scope was intended to be the evaluation path of subschemas, not resources. However, I've always thought that level of granularity was unnecessary and I'd be fine if we all decide that we want to present this concept to people as resources rather than sub-schemas.

At the moment, the only dynamic scope concept we have is dynamic references. That concept only cares about the resource evaluation path because dynamic anchors use a resource-level keyword. I would hope we don't consider adding any more dynamically scoped features, but it's conceivable that a schema-level keyword could could effect dynamically scoped state, in which case the subschema evaluation path becomes relevant. If we want to leave the door open for this kind of thing, we'd want to stick to a subschema evaluation path definition.

### Thoughts on scopes
I think a mistake we commonly make is thinking that some behavior being dependent on the instance is a marker of dynamic scope. For example, the ["multiple dynamic paths"](https://github.com/json-schema-org/JSON-Schema-Test-Suite/blob/main/tests/draft2020-12/dynamicRef.json#L391-L475) test doesn't actually demonstrate dynamic scope. It only shows the need for global state. However, the ["after leaving a dynamic scope, it is not used by a $dynamicRef"](https://github.com/json-schema-org/JSON-Schema-Test-Suite/blob/main/tests/draft2020-12/dynamicRef.json#L478-L515) test does demonstrate dynamic scope as the value of the that global state changes depending on the evaluation path.

### Scope levels
In thinking about scopes, I found it helpful to write out something schema like in a (fake) language that looks more like a programming language where we're more comfortable thinking about scopes. I think this can help demonstrate what I meant about resource-level keywords.

```
// A dynamically scoped global variable that will track the target of the "thingy" dynamic anchor
global thingy

class Main {
  var $id = "https://test.json-schema.org/dynamic-ref-leaving-dynamic-scope/main"
  var $schema = "https://json-schema.org/draft/2020-12/schema"

  main(instance) {
    if (firstScope(instance)) {
      secondScope(instance)
    }
  }
}

class FirstScope {
  var $id = "https://test.json-schema.org/dynamic-ref-leaving-dynamic-scope/first-scope"
  var $schema = "https://json-schema.org/draft/2020-12/schema"

  // Set the value the global "thingy" will have from this point on in the evaluation.
  // This value will be forgotten when evaluation leaves this class
  local thingy = thingy || this.main

  main(instance) {
    type = "number"
  }
}

class SecondScope {
  var $id = "https://test.json-schema.org/dynamic-ref-leaving-dynamic-scope/second-scope"
  var $schema = "https://json-schema.org/draft/2020-12/schema"
  local thingy = thingy || this.main

  main(instance) {
    start(instance)
  }

  isNull(instance) {
    type = "null"
  }
}

class Start {
  var $id = "https://test.json-schema.org/dynamic-ref-leaving-dynamic-scope/start"
  var $schema = "https://json-schema.org/draft/2020-12/schema"

  main(instance) {
    innerScope.anchor(instance)
  }
}

class InnerScope {
  var $id = "https://test.json-schema.org/dynamic-ref-leaving-dynamic-scope/start"
  var $schema = "https://json-schema.org/draft/2020-12/schema"
  local thingy = thingy || this.main

  main(instance) {
    type = "string"
  }

  anchor(instance) {
    thingy(instance)
  }
}
```

The class is like a schema resource and the methods are like subschemas. A class has instance variables are (lexically) in scope for anything in within the class. In a schema, we have a few keywords that behave like this. I call them resource-level keywords. Most keywords would only be visible within their local subschema (or method in this analogy). I call these schema-level keywords.

In this example, the `local` keyword is declaring a dynamically scoped global variable. This is used to set the dynamic anchor. The "thingy" anchor is set at the class level, because dynamic anchors are resource-level. From this example, it might be more clear how it could make sense for someone to define a schema-level keyword that uses dynamic scope. In that case `local` would be used in the method-level rather than the instance-level.

Summary
* Resource-level lexical scope: `$id`, `$schema`, `$anchor`
* Schema-level lexical scope: `type`, `properties`, `if`, `then`, etc.
* Resource-level dynamic scope: `$dynamicAnchor`

mwadams: I think a mistake we commonly make is thinking that some behavior being dependent on the instance is a marker of dynamic scope. For example, the ["multiple dynamic paths"](https://github.com/json-schema-org/JSON-Schema-Test-Suite/blob/main/tests/draft2020-12/dynamicRef.json#L391-L475) test doesn't actually demonstrate dynamic scope. It only shows the need for global state. However, the ["after leaving a dynamic scope, it is not used by a $dynamicRef"](https://github.com/json-schema-org/JSON-Schema-Test-Suite/blob/main/tests/draft2020-12/dynamicRef.json#L478-L515) test does demonstrate dynamic scope as the value of the that global state changes depending on the evaluation path.

This is a very good example: the test I had the single most difficulty with implementing was the "after leaving a dynamic scope, it is not used by a $dynamicRef" one - because it was the only one that meant I really had to understand the mechanism in play.

pages/blog/posts/understanding-lexical-dynamic-scopes.md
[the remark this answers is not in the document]
jdesrosiers: I don't think it makes sense to say that the scope is the entire schema resource. Take this javascript as an analogous scope example.

```javascript
function foo() {
  let a;
  function bar() {
    let b;
    function baz() {
      let c;
    }
  }
}
```

When in the `bar` function, `a` and `b` are in scope. Anything within and lexically above the current block is in scope. If the entire document was in scope, it would mean you would mean `c` is also in scope from `bar`, which I don't think makes sense.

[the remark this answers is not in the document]
jviotti: Tying back to @jdesrosiers 's comment here (https://github.com/json-schema-org/website/pull/274#discussion_r1465740269), if we say a schema resource is analogous to a class and that a subschema is analogous to a method, then maybe what @Relequestual is saying makes sense:

- The lexical scope is the class
- When you are in any of its methods (i.e. subschemas), your scope is indeed the entire class, as you can access its members and other methods

pages/blog/posts/understanding-lexical-dynamic-scopes.md
jviotti: @jdesrosiers I like your points and the programming-language analogy example. I think that's a good mental model for how scoping should work. I particularly like the *schema resource to class* analogy, the *subschema to method* analogy, and the *resource level lexical scope to member* analogy.

> - Resource-level lexical scope: $id, $schema, $anchor
> - Schema-level lexical scope: type, properties, if, then, etc.
> - Resource-level dynamic scope: $dynamicAnchor

I liked what was suggested at some point, which is that both lexical & dynamic scope operate at the subschema level instead of at the resource level (which feels natural too) and that there are some keywords, like `$anchor`, which conceptually affect every lexical scope of the schema resource they are defined in.

jviotti:
@jdesrosiers @gregsdennis @Relequestual Let me attempt to summarize the different points of views I'm seeing in this thread (please correct me if I got any of them wrong, or missed any) for defining lexical & dynamic scopes are the following:

### Proposal 1 (mainly proposed by @gregsdennis and @Relequestual)

- The lexical scope is the entire schema resource of the subschema being evaluated
- The dynamic scope is the ordered set of schema resources visited during the evaluation process at that point

In this model, keywords like `$schema`, `$anchor`, and `$dynamicAnchor` naturally work because the scopes consider the entire schema resource.

### Proposal 2 (mainly proposed by @jdesrosiers)

- The lexical scope is the subschema being evaluated and all of its parent schemas within the same schema resource
  - Notably, the children schemas of a subschema are NOT in the lexical scope
- The dynamic scope is the ordered set of subschemas visited during the evaluation process at that point

In this model, keywords like `$schema`, `$anchor`, and `$dynamicAnchor` work because we consider them to be keywords capable of affecting every scope of the schema resource they are declared in (i.e. they are like members in a class following @jdesrosiers) analogy.

***

Tying back to programming languages like @jdesrosiers suggested, I feel that proposal 1 is more object-oriented (with classes being equal to schema resources) while proposal 2 is more functional (i.e. like the example here: https://github.com/json-schema-org/website/pull/274#discussion_r1466866002).

gregsdennis:
> Tying back to programming languages like @jdesrosiers suggested, I feel that proposal 1 is more object-oriented (with classes being equal to schema resources) while proposal 2 is more functional.

If this isn't an apt analysis of my typical point of view vs Jason's, I don't know what is.

gregsdennis:
I don't mind either of these.

My main point in my original comments was that we should have a practical reason for defining them in a particular way.  If they both work to explain how the associated keywords operate, I'm fine with either.

pages/blog/posts/understanding-lexical-dynamic-scopes.md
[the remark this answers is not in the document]
Relequestual: > I don't think it makes sense to say that the scope is the entire schema resource.

Understood, however an embedded schema resource's contense is not included in the containing schema resource. `baz` cann't see `a` or `b`.

A better analogy is individual functions in separate classes, all imported into the same class, then re-exported. They have their own lexical scope, but can be embedded in other container classes for ease of use.

Relequestual:
@jviotti I think your proposal 2 is incorrect...

> Notably, the children schemas of a subschema are NOT in the lexical scope

You could have an `$anchor` in a subschema, and that would be referenceable in the root schema.

jviotti:
@Relequestual 

> You could have an $anchor in a subschema, and that would be referenceable in the root schema, all within the same schema resource.

Correct, that's why I have a clarification for keywords like `$anchor` in there Jason's suggestion was that `$anchor` is not in scope because the entire schema resource is in scope, but because `$anchor` is on a category of special keywords that always affect its entire schema resource, even if the scope doesn't include it.

pages/blog/posts/understanding-lexical-dynamic-scopes.md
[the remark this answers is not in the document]
jdesrosiers: > When you are in any of its methods (i.e. subschemas), your scope is indeed the entire class, as you can access its members and other methods

Not the entire class. When in a method, you can access your own local variable, but not the local variables of another method. Just like when evaluating `additionalProperties` you only have access to the `properties` keyword in the same subchema, not the `properties` keyword in another subschema. You can access the resource-scoped variables from the entire class, but you can only access method-scope variables from the method you're in.

> A better analogy is individual functions in separate classes, all imported into the same class, then re-exported. They have their own lexical scope, but can be embedded in other container classes for ease of use.

I would compare embedded schemas to inner classes (a class defined inside a class). I'm not sure I entirely follow your analogy, but I think it's effectively the same idea?

jdesrosiers:
> Tying back to programming languages like @jdesrosiers suggested, I feel that proposal 1 is more object-oriented (with classes being equal to schema resources) while proposal 2 is more functional. More notably, in proposal 1, a subschema declaring children subschemas seems to be analogous to a class method calling another class method, while in proposal 2, a subschema declaring children subschemas seems to be analogous to a function defining another function inside (i.e. like the example here: https://github.com/json-schema-org/website/pull/274#discussion_r1466866002).

I don't see any distinction. The scope concepts are the same in both paradigms. The classes/methods analogy describes what I've been saying exactly. I used classes/methods to explain the concepts as I see them. I could have just used functions to demonstrate the same thing, which is what I was doing in the other example with nested functions. There's no difference.

> You could have an `$anchor` in a subschema, and that would be referenceable in the root schema, all within the same schema resource.

Anchors are strange because they're not really lexical or dynamic. They're something else. Maybe you could call them "universal"? It would be like declaring a instance variable inside a method instead of at the class level. The variable would available to the whole class as soon as the class is instantiated whether or not the method ever got called. The best we can do is squint and pretend it was declared in the root of the schema resource like the normal resource-level keywords. If we try to address this problem by saying the entire schema is in scope, we're not taking into account all the other keywords that don't behave that way. When evaluating `additionalProperties`, our scope is limited `properties` keywords only in the same schema, not the entire schema resource.

jviotti:
Post updated! @gregsdennis @jdesrosiers @Relequestual @benjagm Can you take another look? 🙏🏻 

In case it comes handy to some, here is a PDF export of the article: 
[Understanding JSON Schema Lexical and Dynamic Scopes.pdf](https://github.com/json-schema-org/website/files/14196607/Understanding.JSON.Schema.Lexical.and.Dynamic.Scopes.pdf)

pages/blog/posts/understanding-lexical-dynamic-scopes.md
```diff
@@ -0,0 +1,418 @@
...
+
+Lexical Scope
+-------------
+
+Under the graph analogy from the previous section, the lexical scope of a
+schema consists of the node being evaluated. In other words, the lexical scope
+of a schema consists of the entire schema resource of which it is part of. 
+
+Consider the following sequence of examples. In the left, a JSON Schema with a
+single nested schema resource. In the right, the corresponding directed graph
```
gregsdennis, line 104:
```suggestion
single nested schema resource. On the right, the corresponding directed graph
```

We typically use "on" when there's only one direction like "top" or "left".

pages/blog/posts/understanding-lexical-dynamic-scopes.md
```diff
@@ -0,0 +1,418 @@
...
+Schemas as Directed Graphs
+--------------------------
+
+JSON Schema is a recursive data structure. In the context of schema resources,
+this means that a schema resource may introduce nested schema resources (like
+we saw on the previous section) and use referencing keywords (like `$ref`) to
+point to external schema resources, effectively creating a directed graph of
+schema resources. 
+
+Consider the following example. In the top left, a root schema resource named
```
gregsdennis, line 81: However, we do use "in" when there are multiple directions, like "top right" or "bottom left".

English is dumb.

pages/blog/posts/understanding-lexical-dynamic-scopes.md
```diff
@@ -0,0 +1,418 @@
...
+
+As you will see, thinking of a schema as a directed graph of schema resources
+greatly helps in understanding both lexical and dynamic scopes.
+
+Lexical Scope
+-------------
+
+Under the graph analogy from the previous section, the lexical scope of a
+schema consists of the node being evaluated. In other words, the lexical scope
+of a schema consists of the entire schema resource of which it is part of.
```
gregsdennis, line 101:
```suggestion
of a schema consists of the entire schema resource to which it belongs. 
```

pages/blog/posts/understanding-lexical-dynamic-scopes.md
```diff
@@ -0,0 +1,418 @@
...
+
+### Following References
+
+When the evaluation process encounters a reference keyword, it *abandons* the
+lexical scope of the reference schema and enters the lexical scope of the
+*destination* schema.
+
+If the reference points to a subschema *within* the same schema resource, the
+lexical scope effectively remains the same. Coming back to the graph analogy,
+the evaluation process remains at the same node. However, if the reference
```
gregsdennis, line 157: > Coming back to the graph analogy, the evaluation process remains at the same node.

This isn't right.  The evaluation process always changes _nodes_ by following the `$ref`.  It's just that in this case, the lexical scope remains the same.

Your image is correct: the lexical scope remains the same, but the evaluated node changes from `/items` to `/$defs/person-name`.

pages/blog/posts/understanding-lexical-dynamic-scopes.md
```diff
@@ -0,0 +1,418 @@
...
+corresponding directed graph representations of the root schema resource, the
+nested schema resource, and the external schema resource. Like before, at each
+step of the evaluation process, we gray out the parts of the schema and of the
+directed graph that are not part of the lexical scope.
+
+The evaluation process starts with the top level schema. The lexical scope at
+that point is the root schema resource, and both the nested schema resource and
+the external schema resource are out of scope: 
+
+![Lexical scope after following a reference accross resources (1)](/img/posts/2024/understanding-lexical-dynamic-scopes/lexical-scope-ref-remote-1.webp)
```
gregsdennis, line 185: The text shows it correctly, but the tree representation in the image seems to convey that the `anyOf` subschemas (the ones with the `$ref`s) aren't part of the top-level lexical scope, but they are.  I'm not sure how to convey that in the tree.

Probably not a big deal.

pages/blog/posts/understanding-lexical-dynamic-scopes.md
```diff
@@ -0,0 +1,418 @@
...
+
+In comparison to the lexical scope, the dynamic scope of a schema cannot always
+be statically determined, as the evaluation path often depends on the instance.
+For example, for schemas that make use of logic applicator keywords such as
+`if` or `oneOf`, the ordered sequence of schema resources in scope may vary
+depending on the characteristics of the instance.
+
+### Following References
+
+So far, we learnt that for the lexical scope, following a reference consists in
```
gregsdennis, line 245: Are we going with American English for posts?  If so, then "learnt" should be "learned".

Also a couple other suggestions.

```suggestion
So far, we've learned that for the lexical scope, following a reference consists of
```

pages/blog/posts/understanding-lexical-dynamic-scopes.md
```diff
@@ -0,0 +1,418 @@
...
+`https://example.com/item` (from `/$defs/person/$ref`) that is part of an
+external schema resource called `https://example.com/people` shown in the
+bottom left. In the right, a directed graph representation of the relationship
+between these schema resources and the dynamic scope.
+
+Like the other examples so far, the evaluation process starts with the top
+level schema. The dynamic scope at that point is the root schema resource, and
+all other schema resources are out of scope:
+
+![The dynamic scope and remote references (1)](/img/posts/2024/understanding-lexical-dynamic-scopes/dynamic-scope-ref-remote-1.webp)
```
gregsdennis, line 281: > `"$ref": "people/#items"

We usually advise against referencing directly into other keywords like this.  The advice is generally to extract the target into `$defs` or to an external schema and reference from both locations.

I recognize that you have limited space here, though.

gregsdennis: Also since that subschema defines an `$id`, this is using a pointer reference across resource boundaries, which is undefined.  (Most implementations support it, but it's not a requirement to support.)

The target subschema here has a resolved ID of `https://example.com/item` and the reference can just be changed to `"$ref": "item"`, which feels weird.

pages/blog/posts/understanding-lexical-dynamic-scopes.md
```diff
@@ -0,0 +1,418 @@
...
+
+Understanding how static and dynamic scopes work is essential for gaining a
+deeper understanding of JSON Schema. The most important points to remember are
+summarized in the following table:
+
+| Comparison Point                     | Lexical Scope | Dynamic Scope |
+|--------------------------------------|---------------|---------------|
+| **Definition**                       | Consists of the schema resource being evaluated | Consists of the stack of schema resources evaluated so far |
+| **Determining the scope**            | Can be statically determined without taking instances into account | Cannot be always statically determined. It may vary depending on the instance |
+| **Following references**             | Consists in abandoning the lexical scope of the origin schema and entering the lexical scope of the destination schema | Consists in pushing the destination schema resource to the top of the dynamic scope stack |
```
gregsdennis, line 403:
```suggestion
| **Following references**             | Consists of abandoning the lexical scope of the origin schema and entering the lexical scope of the destination schema | Consists of pushing the destination schema resource to the top of the dynamic scope stack |
```

gregsdennis (changes requested):
Overall _very_ good.  Significant improvement over the previous.

Still a few questions/comments.

pages/blog/posts/understanding-lexical-dynamic-scopes.md
```diff
@@ -0,0 +1,418 @@
...
+
+In this case, we only push the destination schema resource to the dynamic
+scope, *jumping* over the root schema resource of the external schema.  After
+following this final reference, the dynamic scope now consists of
+`https://example.com` (the root schema resource), followed by
+`https://example.com/name` (the nested schema resource at `/properties/name`),
+followed by `https://example.com/person` (the nested schema resource at
+`/$defs/person`), followed by `https://example.com/item` (the nested schema
+resource of the external schema). Note that the root schema resource called
+`https://example.com/people` is not part of the dynamic scope:
```
jdesrosiers, line 316: I don't think that's correct. The reference is `people#/items`. Evaluation has to get `https://example.com/people` before it can follow the pointer (`/items`), and only then does it move into `https://example.com/item`. I think `https://example.com/people` is in the dynamic scope because you had to pass through it to get to `https://example.com/item`. At least that's the way my implementation works.

However, if the reference was just `items`, it would go directly to `https://example.com/items` without passing through `https://example.com/people` first. Although I'm not sure you can count on implementations being able directly resolve to an embedded schema in a completely separate schema document, so I'm not sure just changing it to `items` is a great solution.

jdesrosiers (commented):
This is looking great so far! I have a couple of suggestions inline.

pages/blog/posts/understanding-lexical-dynamic-scopes.md
Relequestual: Fixed

pages/blog/posts/understanding-lexical-dynamic-scopes.md
```diff
@@ -0,0 +1,418 @@
+---
+title: "Understanding JSON Schema Lexical and Dynamic Scopes"
+date: "2024-02-07"
+tags:
+  - Fundamentals
+  - Scopes
+type: Engineering
+cover: /img/posts/2024/understanding-lexical-dynamic-scopes/cover.webp
```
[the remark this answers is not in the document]
jviotti: My bad. Seems like I accidentally removed it with my other image updates yesterday.

pages/blog/posts/understanding-lexical-dynamic-scopes.md
jviotti: I guess my English is completely mixed up across regions at this point 😅

pages/blog/posts/understanding-lexical-dynamic-scopes.md
gregsdennis: @jviotti there are several of these.  I didn't want to mark them all.

jviotti:
Thanks for the new comments! I addressed the simple grammar ones and plan to address the other ones tomorrow.

pages/blog/posts/understanding-lexical-dynamic-scopes.md
jviotti: Ah, got it. Fixing them all now...

pages/blog/posts/understanding-lexical-dynamic-scopes.md
jviotti: We are considering nodes to be schema resources. If we take into account references within a single schema resource, the reference changes subschemas, but not schema resources, so the node (at least how we define) remains the same, right?

pages/blog/posts/understanding-lexical-dynamic-scopes.md
jviotti: I see your point. The labels in the arrow were meant to convey what reference is followed when following that arrow, and not whether they are part of the scope or not.

If you think its confusing, maybe worth getting rid of the arrow labels?

pages/blog/posts/understanding-lexical-dynamic-scopes.md
jviotti: Would it make it nicer if I set an anchor at `item` and then reference i.e. `people#item`?

pages/blog/posts/understanding-lexical-dynamic-scopes.md
```diff
@@ -0,0 +1,418 @@
...
+    byline: O'Reilly author, founder of Intelligence.AI and Sourcemeta
+excerpt: "A deep dive on lexical and dynamic scopes, essential for understanding more advanced topics like dynamic referencing"
+---
+
+Most of the [keywords](https://json-schema.org/learn/glossary#keyword) defined
+by the JSON Schema organization can be either evaluated on their own, or by
+only considering the values of their adjacent keywords. For example, the `type`
+keyword is independent of any other keywords, while the `minContains` and
+`maxContains` keywords depend on the `contains` keyword defined in the same
+schema object.
```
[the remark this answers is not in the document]
jviotti: Makes sense. On it...

pages/blog/posts/understanding-lexical-dynamic-scopes.md
```diff
@@ -0,0 +1,418 @@
...
+Note that by definition, the lexical scope of any subschema can be statically
+determined without taking
+[instances](https://json-schema.org/learn/glossary#instance) into account, just
+as we did here.
+
+### Lexical Scope and Anchors
+
+As another practical example, consider the `$anchor` keyword that defines a
+[location-independent
+identifier](https://json-schema.org/draft/2020-12/json-schema-core#section-8.2.2)
```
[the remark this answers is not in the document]
jviotti: It is, thanks!

pages/blog/posts/understanding-lexical-dynamic-scopes.md
```diff
@@ -0,0 +1,418 @@
...
+`https://example.com/origin` that declares a nested schema resource named
+`https://example.com/nested` (at `/properties/bar`) and references an external
+schema resource named `https://example.com/destination` (from
+`/properties/foo/$ref`). In the bottom left, a root schema resource named
+`https://example.com/destination` that references a nested schema resource
+called `https://example.com/nested-string` (from `/items/$ref`). In the right,
+a directed graph representation of the relationship between these schema
+resources:
+
+![Thinking of a JSON Schema as a directed graph](/img/posts/2024/understanding-lexical-dynamic-scopes/directed-graph.webp)
```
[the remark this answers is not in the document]
jviotti: Ah, true. Good catch! On it

pages/blog/posts/understanding-lexical-dynamic-scopes.md
jviotti: From what I understood from other discussions, most implementation consist of two steps:

- Scanning the given schema to find identifiers, embedded schema resources, anchors, etc
- Evaluating the given schema

I agree that `https://example.com/people` is *scanned*, but I understood that it is not *evaluated*. Therefore, it is not part of the dynamic scope. 

In any case, I think this is a good point worth clarifying more in the post once we align. @gregsdennis @Relequestual @Julian What is your take on this one?

pages/blog/posts/understanding-lexical-dynamic-scopes.md
gregsdennis: Oh, I see: "node" refers to the graph here.  That wasn't clear to me at first.  I was thinking "JSON node".  I think your rewording makes that clearer. 👍

pages/blog/posts/understanding-lexical-dynamic-scopes.md
gregsdennis: Perhaps that.  The only other thing I can think of is to expand the dotted boundary to encompass the labels.  I'll let you decide.

pages/blog/posts/understanding-lexical-dynamic-scopes.md
gregsdennis: If you do that, I think you lose your example where you're skipping over `people`.

Will continue to follow up in the [other thread](https://github.com/json-schema-org/website/pull/274#discussion_r1482235743).

pages/blog/posts/understanding-lexical-dynamic-scopes.md
gregsdennis: > I agree that `https://example.com/people` is _scanned_, but I understood that it is not _evaluated_.

Yes, that's true.  However, it requires that an implementation scans the schema when it's registered, which may not be the case.  If it doesn't then it doesn't know where to find `items` when it's referenced.

I agree with @jdesrosiers in that with the `people#/items` reference, it has to _get_ `people` first, but I don't think it would be part of the dynamic scope since `items` is a new resource.

> In any case, I think this is a good point worth clarifying more in the post once we align.

This is a sticky nuance that I don't think impacts what the post is trying to say, and it's probably best avoided for this post.  I think we'd be better off with a different kind of example.

(That said, I would like to see some test cases for this to see what different implementations do.  I'll work on that separately.)

My argument against this particular example is that it promotes having a pointer reference cross resource boundary, which is something the spec advises against.

gregsdennis: I've opened an issue in the test suite to have a case added, but I also think the schema I have there may help this example a bit since it removes the pointer-crossing-resource-boundary issue.

https://github.com/json-schema-org/JSON-Schema-Test-Suite/issues/719

pages/blog/posts/understanding-lexical-dynamic-scopes.md
jviotti: Great! Let me clarify it a bit more on the post anyway

pages/blog/posts/understanding-lexical-dynamic-scopes.md
jviotti: I removed the labels!

pages/blog/posts/understanding-lexical-dynamic-scopes.md
jviotti: I think its OK, as even with `people#item`, `people` is never at the dynamic scope, right?

jviotti: Right. Let's continue the discussion on the other thread :+1:

pages/blog/posts/understanding-lexical-dynamic-scopes.md
jviotti: Makes sense. Let me simplify this example to remove the pointer reference cross resource boundary. We can avoid touching on this edge case

jviotti: Done! Please take a look at the revised example

jviotti:
OK, I addressed all the comments so far. Please give it another look!

pages/blog/posts/understanding-lexical-dynamic-scopes.md
```diff
@@ -0,0 +1,412 @@
...
+examples so far only considered *pushing* schema resources to the top the of
+stack. 
+
+In traditional programming languages, program execution typically involves
+procedures calling other procedures, creating what is referred to in Computer
+Science as a [call stack](https://en.wikipedia.org/wiki/Call_stack).
+Eventually, a procedures will not call any other procedures. When such leaf
+procedures finish executing, the call stack will
+[unwind](https://en.wikipedia.org/wiki/Call_stack#Unwinding) (a pop operation)
+and control will return the caller frame.
```
jdesrosiers, line 327:
```suggestion
In traditional programming languages, program execution typically involves
procedures calling other procedures, creating what is referred to in Computer
Science as a [call stack](https://en.wikipedia.org/wiki/Call_stack).
Eventually, a procedure will not call any other procedures. When such leaf
procedures finish executing, the call stack will
[unwind](https://en.wikipedia.org/wiki/Call_stack#Unwinding) (a pop operation)
and control will return to the caller frame.
```

pages/blog/posts/understanding-lexical-dynamic-scopes.md
```diff
@@ -0,0 +1,412 @@
...
+-------
+
+Understanding how static and dynamic scopes work is essential for gaining a
+deeper understanding of JSON Schema. The most important points to remember are
+summarized in the following table:
+
+| Comparison Point                     | Lexical Scope | Dynamic Scope |
+|--------------------------------------|---------------|---------------|
+| **Definition**                       | Consists of the schema resource being evaluated | Consists of the stack of schema resources evaluated so far |
+| **Determining the scope**            | Can be statically determined without taking instances into account | Cannot be always statically determined. It may vary depending on the instance |
```
jdesrosiers, line 396:
```suggestion
| **Determining the scope**            | Can be statically determined without taking instances into account | Cannot always be statically determined. It may vary depending on the instance |
```

jdesrosiers (approved):
Looks great! Thanks for putting in all the hard work on this. I'm sure it ended up being more of a can of worms than you expected.

jviotti:
> I'm sure it ended up being more of a can of worms than you expected.

Haha it's good though. My main objective with this was improving my understanding of this whole thing, and that's definitely checked!

gregsdennis (approved):
Looks great!  Thanks for putting all this together and for putting up with us.

jviotti:
@Relequestual I think you are the one missing here. Also, how do we manage publication date? I guess I can bump it up to the correct date once its approved?

Relequestual (approved):
I very much appreciate you sticking with us on this and working through all the feedback and the can of worms we opened with this!

Great work, and we should be super excited to have concluded this and have it ready for publishing! 🥳

benjagm (approved):
Thanks for the great work Juan!!

jviotti:
Thank you all again for the great reviews and for patiently helping me polish my understanding of this! 🙏🏻 @benjagm Let's coordinate the release.

PR #277, opened 2024-01-24
add glossary entry for 'implementation'

gregsdennis:
**GitHub Issue:**

Resolves https://github.com/json-schema-org/json-schema-spec/issues/1440

**Summary**:

Adds a glossary entry for 'implementation' as discussed in the issue above.

No

### DIFF
--- a/pages/learn/glossary.md
+++ b/pages/learn/glossary.md
@@ -55,7 +55,7 @@ The [JSON Schema specification](../specification) defines behavior for a large l
 
 ### implementation
 
-Historically, the word "implementation" has been used to describe both specifically validators/annotators (interpreted as strictly operating as the spec describes) and also various other kinds of JSON Schema tooling.  However, due to this ambiguity, we have decided that the preferred conversational term should be ["tooling"](#tooling).
+Historically, the word "implementation" has been used to describe both specifically validators/annotators (the scope of the spec) and also various other kinds of JSON Schema tooling.  However, due to this ambiguity, we have decided that the preferred conversational term should be ["tooling"](#tooling).
 
 ### instance
 
--- a/pages/learn/glossary.md
+++ b/pages/learn/glossary.md
@@ -55,7 +55,7 @@ The [JSON Schema specification](../specification) defines behavior for a large l
 
 ### implementation
 
-An implementation is any piece of software that provides or exposes behavior defined by the JSON Schema specification.
+An implementation is any software tooling or library for working with or evaluating schemas in some way that does not contradict the JSON Schema specification.
 
 The following are considered implementations:
 
--- a/pages/learn/glossary.md
+++ b/pages/learn/glossary.md
@@ -53,6 +53,19 @@ A property appearing within a [schema](#schema) object.
 
 The [JSON Schema specification](../specification) defines behavior for a large library of keywords which can be used to describe [instances](#instance).
 
+### implementation
+
+An implementation is any piece of software that provides or exposes behavior defined by the JSON Schema specification.
+
+The following are considered implementations:
+
+- a validator library
+- a CLI validator
+- a schema or code generator
+- a UI form generator
+
+Applications which use JSON Schema internally without exposing that functionality in some way, for example, validating configuration files or web requests, are not considered implementations.
+
 ### instance
 
 A piece of JSON data which is to be described by a [schema](#schema).
--- a/pages/learn/glossary.md
+++ b/pages/learn/glossary.md
@@ -55,16 +55,7 @@ The [JSON Schema specification](../specification) defines behavior for a large l
 
 ### implementation
 
-An implementation is any software tooling or library for working with or evaluating schemas in some way that does not contradict the JSON Schema specification.
-
-The following are considered implementations:
-
-- a validator library
-- a CLI validator
-- a schema or code generator
-- a UI form generator
-
-Applications which use JSON Schema internally without exposing that functionality in some way, for example, validating configuration files or web requests, are not considered implementations.
+Historically, the word "implementation" has been used to describe both specifically validators/annotators (interpreted as strictly operating as the spec describes) and also various other kinds of JSON Schema tooling.  However, due to this ambiguity, we have decided that the preferred conversational term should be ["tooling"](#tooling).
 
 ### instance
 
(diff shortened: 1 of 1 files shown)


Relequestual (changes requested):
I think this is at odds with itself.
Code generation doesn't provide or expose anything defined in the specification.
It interprets the semantics of what's defined in the specification.

> Applications which use JSON Schema internally without exposing that functionality in some way, for example, validating configuration files or web requests, are not considered implementations.

We list VSCode on the implementations page.

I think this is well intentioned, but this definition doesn't quite work.

jdesrosiers (approved):
This sounds great to me!

jdesrosiers:
@Relequestual 

> Code generation doesn't provide or expose anything defined in the specification.

I agree that these kinds of things are a bit of a stretch, but they certainly do implement aspects of the spec. They use the structure of the schema, use the same keywords, and implement identification and referencing as defined in the spec. If there's a better way to express this, I'm for it, but I don't think it's inconsistent. There are certainly some grey areas, but I think this definition successfully gets the right idea across.

> > Applications which use JSON Schema internally without exposing that functionality in some way, for example, validating configuration files or web requests, are not considered implementations.
>
> We list VSCode on the implementations page.

I think you're misunderstanding what that line is saying. VSCode's fits that definition just fine because the functionality is exposed. JSON Schema based validation is presented to the user as inline error messages (as well as several other user facing features). What's not considered an implementation is an application like a REST API that accepts requests and validates those requests against a JSON Schema. The user may never know or care that JSON Schema was used under the hood to validate their request.

pages/learn/glossary.md
```diff
@@ -53,6 +53,19 @@ A property appearing within a [schema](#schema) object.
 
 The [JSON Schema specification](../specification) defines behavior for a large library of keywords which can be used to describe [instances](#instance).
 
+### implementation
+
+An implementation is any piece of software that provides or exposes behavior defined by the JSON Schema specification.
```
jdesrosiers, line 58: Taking a stab at addressing @Relequestual's concerns about the wording here...

```suggestion
An implementation is any software tooling or library for working with or evaluating schemas in some way that is not inconsistent with the JSON Schema specification.
```

gregsdennis: "not inconsistent"?

not just "consistent"?

jdesrosiers: I thought about that, but I thought it was more clear this way despite the double negative. I wanted to convey that an implementation might do more or less than what the spec says as long as as it doesn't contradict it. I felt like "consistent" could be interpreted as referring to the whole spec, which is not necessarily the case for all types of implementations.

But, I'm bad at words, so that might not make any sense.

Relequestual (changes requested):
I 100% agree that this is a better definition, more in line with what we expect (Thanks @jdesrosiers). However, when I try to apply this to our current list, it raises some problems.

Consider: AJV. We know the default behaviour is non-compliant. Isn't any "non-compliance" with the spec (aka failing any test), legalistically in contradiction of the spec?

Unless anyone has a clear and obvious solution, maybe this could be discussed at our next OCWM?

mwadams:
> I 100% agree that this is a better definition, more in line with what we expect (Thanks @jdesrosiers). However, when I try to apply this to our current list, it raises some problems.
> 
> Consider: AJV. We know the default behaviour is non-compliant. Isn't any "non-compliance" with the spec (aka failing any test), legalistically in contradiction of the spec?
> 
> Unless anyone has a clear and obvious solution, maybe this could be discussed at our next OCWM?

I think this is possibly just trying to conflate two things:

1. Do I *try* to implement (some version or versions) of the spec
2. To what extent is that implementation *compliant* with the spec

So we have _implementations_ with specific _compliance_.

Julian:
This has gone past lots of people at this point, and really has had more discussion than any glossary entry probably deserves :D -- so I hesitate to throw in another opinion. So I caveat that I don't think it matters much, we can always change things later if need be, so feel free to ignore my comment. Now, caveat aside:

This seems odd to me. I think to me this is "misusing" the glossary, which I think is why the definition seems awkward.

Glossaries are (I assert) intended to help unfamiliar users [^1] understand how an *already used term* is *already used* in a quick and self-contained way. The balance should be on colloquial use, with when possible a precise definition, but if the precise one isn't possible (because it relies on too much other knowledge) then fine, just the colloquial one.

To me this doesn't really do that -- to me what this does is document to *us ourselves* (the internal team) what we *want* the definition to be, and it does so for some complete side purpose -- namely defining what software we want to put on a page on the site which *happens* to be titled "Implementations" but could equally well be titled "JSON Schema Software" or "Stuff We Like".

If it were me, I would have written something like:

> a concrete piece of software which exposes functionality defined by the JSON Schema specification to end users. <some of the examples listed here> Often -- but not always -- an implementation is synonymous with "data validator" (n.b. which I believe is the case)

and left it there. That indeed may exclude "code generation" -- i.e. taking a schema and generating a class in your programming language -- from being an implementation. I personally indeed would not call such software an "implementation of JSON Schema", even if it's software using JSON Schema. And I think that matches how people would colloquially talk about such software. And again I don't think that should have any relevance to whether we list that software on the page on the site :)

Just my 0.02, again feel free to take or leave.

[^1]: a beginner, or anyone who doesn't already know what a specific term means

Relequestual (approved):
Thanks for revising. I think this accurately reflects what we agree on, while also being easy enough to understand for the intended audience.

pages/learn/glossary.md
```diff
@@ -53,6 +53,10 @@ A property appearing within a [schema](#schema) object.
 
 The [JSON Schema specification](../specification) defines behavior for a large library of keywords which can be used to describe [instances](#instance).
 
+### implementation
+
+Historically, the word "implementation" has been used to describe both specifically validators/annotators (interpreted as strictly operating as the spec describes) and also various other kinds of JSON Schema tooling.  However, due to this ambiguity, we have decided that the preferred conversational term should be ["tooling"](#tooling).
```
jdesrosiers, line 58:
```suggestion
Historically, the word "implementation" has been used to describe both specifically validators/annotators and also various other kinds of JSON Schema tooling.  However, due to this ambiguity, we have decided that the preferred conversational term should be ["tooling"](#tooling).
```

pages/learn/glossary.md
```diff
@@ -112,6 +116,19 @@ Said more plainly, whether a particular value is a subschema or not depends on i
 
 Subschemas may themselves contain sub-subschemas, though colloquially one generally uses the term "subschema" regardless of the level of nesting, further clarifying which larger schema is the parent schema whenever needed.
 
+### tooling
+
+A JSON Schema tool (or colloquially "tooling") is any software application or library for working with or evaluating schemas in some way that does not contradict the JSON Schema specification.
```
jdesrosiers, line 121:
```suggestion
A JSON Schema tool (or colloquially "tooling") is any software application or library for working with or evaluating schemas in some way.
```

jdesrosiers (commented):
Perhaps we don't need to mention the spec in these definitions. I think @Relequestual had a good point about ajv which intentionally violates that spec, but we still consider it JSON Schema tooling.

Suggested changes are inline if people agree.

pages/learn/glossary.md
gregsdennis: I think that note is important for context.

jdesrosiers: I'm not suggesting the historical note be removed. My concern is that this sounds like something is not considered an implementation if it doesn't strictly conform to the spec. It sounds like ajv wouldn't be considered an implementation, which isn't accurate, even historically.

gregsdennis: Okay.  The note you're suggesting to remove is intended to express that the spec only covers validation/annotation.  What would you have to express that?

jdesrosiers: My best suggestion might be "(which is what the spec describes)". The words feel a little awkward, but I think it conveys the intended meaning.

gregsdennis: How now?

jdesrosiers: Perfect :+1:

PR #285, opened 2024-01-31
Manfred Case Study

benjagm:
**GitHub Issue:** # / NA

**Summary**: This is the Spanish version of Manfred's Case Study. Already reviewed by @Relequestual and Manfred's team using google docs.

No


marcraminv (commented):
LGTM! Un par de typos y nada más 👍

Relequestual:
Screenshot 2024-02-02 at 13 00 39

This is not ideal in terms of the text over the image.
We previously added drop shadow to the text to help with image colour, but that doesn't help when the image also has text.

I suggest making the parent div of the image having a black background, and the image itself having an opacity of 0.7. Should be able to do this with tailwind as opposed to custom css.

benjagm:
> I suggest making the parent div of the image having a black background, and the image itself having an opacity of 0.7. Should be able to do this with tailwind as opposed to custom css.

Thanks for the specific suggestion. I implemented it and now looks much better. Please check the last PR:
https://cfa353ec.website-2v2.pages.dev/blog

PR #288, opened 2024-02-05
fix: remove the `.idea` directory and added in thew `gitignore`

### LINKED ISSUE #283: Remove the `.idea` directory and add it to .gitignore file
> Is the whole `.idea` directory something controlled by Intelij?
> I think we should probably add it to the `.gitignore` file.
> (Not in this PR, just in general. A new PR.)

_Originally posted by @Relequestual in https://github.com/json-schema-org/website/pull/276#discussion_r1466224686_

Min2who:
Fixes #283


Relequestual (approved):
Thanks for picking up this task.
Houses are built with many individual bricks!

PR #290, opened 2024-02-05
fix: added linting in the workflow

Min2who:
**GitHub Issue:** #289 

**Summary**:

Added linting of pr in every pr

**Do you think resolving this issue might require an [Architectural Decision Record (ADR)](https://github.com/json-schema-org/community/blob/main/CONTRIBUTING.md#key-architectural-decisions)? (significant or noteworthy)**

No
[JUSTIFICATION]

PR #296, opened 2024-02-08
ci: added linting ci in the repo

### LINKED ISSUE #289: Improve the CI/CD pipeline to add linting in the repository
As there is no workflow for linting in the repository, it will be really helpful

Min2who:
Fixes #289 

follow up to #290 

tested locally [here](https://github.com/Min2who/website/actions/runs/7832403448/job/21371015043?pr=1)

image

### DIFF
--- a/.github/workflows/lint.yml
+++ b/.github/workflows/lint.yml
@@ -19,6 +19,3 @@ jobs:
 
       - name: Run linting check
         run: npm run lint
-
-      - name: Check formatting
-        run: npm run lint-fix
--- a/.github/workflows/lint.yml
+++ b/.github/workflows/lint.yml
@@ -1,0 +1,24 @@
+name: Linting on PR
+on:
+  pull_request:
+    branches:
+      - '**'
+
+jobs:
+
+  Continuous-Integration:
+
+    name: Performs linting on the application
+    runs-on: ubuntu-latest
+    steps:
+      - name: Checkout the Repository
+        uses: actions/checkout@v4
+
+      - name: Install Dependencies
+        run: npm install --legacy-peer-deps
+
+      - name: Run linting check
+        run: npm run lint
+
+      - name: Check formatting
+        run: npm run lint-fix


Min2who:
I think we can merge this @Relequestual

Relequestual (commented):
Thanks for this.
I'm not sure about running lint-fix.
What will that do in the context of a GitHub action?

Have you confirmed that it will show any errors? =]

Min2who:
The lint-fix script likely runs a linter with an automatic fix option enabled. In the context of a GitHub Action, running `npm run lint-fix` as mentioned in the script will execute a command that checks your code for style or syntax errors and attempts to automatically fix them where possible.

cc @Relequestual

Relequestual:
> In the context of a GitHub Action, running `npm run lint-fix` as mentioned in the script will execute a command that checks your code for style or syntax errors and attempts to automatically fix them where possible.

It will fix the code it has

Min2who:
> It may fix the code it has, but it won't commit it back to the repo. When the PR is from a fork, it won't have permissions to commit in that fork. When the PR is not from a fork, but a local branch, you could commit, but this currently does not.
> 
Thanks for pointing that out. You're right; the `lint-fix` script may automatically fix code issues but won't commit them back to the repository.

> I don't think having autofix here makes sense. Open to other opinions.

 I agree that having autofix in this context might not be the best approach.

> I meant, have you confirmed that when there are linting errors, they are shown in the action output and make the check fail?

If there are linting issues , the `lint` scripts will cause the workflow to fail since it will return a non-zero exit code for errors

benjagm:
@Min2who I think this functionality is going to be covered by this PR #335 right @aialok?

Min2who:
> @Min2who I think this functionality is going to be covered by this PR #335 right @aialok?

we should close this issue #289 and this PR too

aialok:
> @Min2who I think this functionality is going to be covered by this PR #335 right @aialok?

@benjagm 
No, the functionality in #335 and the functionality in this issue are completely different. Both are performing the same task, but they serve different purposes. This functionality adds the same checks, but in the Continuous Integration Pipeline, specifically in our GitHub Actions workflow, whereas the husky one will be a local check only. The husky one will check these conditions only when we try to commit locally.

There are cases where contributors bypass husky checks somehow. This is why we need to set up a CI pipeline in GitHub Actions for each PR to review whether it has passed the checks or not. This is crucial for the overall stability and safety of the project.

I have already discussed  in this issue: https://github.com/json-schema-org/website/issues/340#issuecomment-1961270454

This issue and #340 are addressing the same things, but #340 also includes checking for linting and type checking.

PR #318, opened 2024-02-20
docs: fix type reference markdown tables

smoya:
**Summary**:

Some of the markdown tables in the [reference guide for types](https://json-schema.org/understanding-json-schema/reference/type) are broken. In particular, both Objective-C and Swift. That ends up being rendered wrongly:

Google Chrome_1l1C24MR

This PR fixes both tables and also reformats the others so they all follow the same format.

### DIFF
--- a/pages/understanding-json-schema/reference/type.md
+++ b/pages/understanding-json-schema/reference/type.md
@@ -26,14 +26,14 @@ go by different names.
 The following table maps from the names of JSON types to their
 analogous types in Python:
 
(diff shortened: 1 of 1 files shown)


smoya:
BTW, Footnotes support in Github's Markdown is a thing https://github.blog/changelog/2021-09-30-footnotes-now-supported-in-markdown-fields/

Relequestual (approved):
Thank you for your contribution Sergio! Much appreciated!

Issue #320, opened 2024-02-20
Rename Implementations by Tools


benjagm:
#### Reason/Context
This ha been decided in [Open Community Working Meeting 2024-02-05 - 14:00 PT](https://github.com/json-schema-org/community/issues/597)

#### Description
We are renaming the Implementations section in favour of Tools

**Do you think this work might require an [Architectural Decision Record (ADR)](https://github.com/json-schema-org/community/blob/main/CONTRIBUTING.md#key-architectural-decisions)? (significant or noteworthy)**

No

PR #321, opened 2024-02-20
Rename implementations section to tools

benjagm:
**GitHub Issue:** #320 
Resolves: #320

**Summary**: This PR renames the Implementations section in favour of Tools

No

### DIFF
--- a/components/Layout.tsx
+++ b/components/Layout.tsx
@@ -122,8 +122,8 @@ const MainNavigation = () => {
       <MainNavLink
         className='hidden lg:block hover:underline'
         uri='/implementations'
-        label='Implementations'
-        isActive={section === 'implementations'}
+        label='Tools'
+        isActive={section === 'tools'}
 
       />
       <MainNavLink
@@ -185,9 +185,9 @@ const MobileNav = () => {
       />
 
       <MainNavLink
-        uri='/implementations'
-        label='Implementations'
-        isActive={section === 'implementations'}
+        uri='/tools'
+        label='Tools'
+        isActive={section === 'tools'}
 
       />
       <MainNavLink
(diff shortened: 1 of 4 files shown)


Relequestual:
Thanks!
How do we plan to handle redirects?

benjagm:
> Thanks! How do we plan to handle redirects?

The url is the same. I just renamed the title/buttons.

Relequestual:
We should also do https://json-schema.org/obsolete-implementations
Happy if you want that to be a separate PR, but I think it could be in this one.

benjagm:
> We should also do https://json-schema.org/obsolete-implementations
> Happy if you want that to be a separate PR, but I think it could be in this one.

I just pushed the change to rename that page as well.

Relequestual:
OK. I'm fine with this if we also create a new good first issue to clean up the rest of the site to use "tools" rather than "implementations" everywhere but the spec documents (because we don't change those. I think that would accurately reflect what we want to see.

And in the code, to make it less confusing.

gregsdennis:
> everywhere but the spec documents (because we don't change those.

Also consider that the spec _does_ need to define its term for "implementation".  I still think that it needs to be "software that performs the tasks defined in the spec", which are just validation and annotation.

praveen-rikhari:
Hello @benjagm I was going through the preview for this branch but I still see that we have implementations term being used in the body of the Tools at a lot of places, I was wandering, should that be done in a different PR??

praveen-rikhari:
Screenshot from 2024-02-22 12-17-38
Screenshot from 2024-02-22 12-17-48
Screenshot from 2024-02-22 12-18-00
 
Here are some of them

Relequestual:
@praveen-rikhari Please reference this issue when you do. Thanks.

PR #323, opened 2024-02-22
WIP: add warning about the use of oneof with recursive references

unguul:
add warning about the use of `oneof` with recursive references so people don't panic when processing times increase 10x from a seemingly small change

**GitHub Issue:** https://github.com/json-schema-org/community/issues/520 & https://github.com/json-schema-org/understanding-json-schema/pull/131

**Summary**:

Using `oneOf` entries with recursive references or simply a high number of references results in long processing times which is actually expected given the nature of `oneOf` (i.e. it has to check *every* referenced schema). However, this is not immediately obvious during development.

No, because it's just a tiny warning added to the docs.

### DIFF
--- a/pages/understanding-json-schema/reference/combining.md
+++ b/pages/understanding-json-schema/reference/combining.md
@@ -135,6 +135,9 @@ Multiple of *both* 5 and 3 is rejected.
 15
 ```
 
+.. warning::
+   Careful consideration should be taken when using `oneOf` entries as the nature of it requires verification of all sub-schemas so doing anything recursive can have a severe performance impact.
+
 <Keywords label="single: not single: schema composition; not" />
 
 ## not


Relequestual (approved):

Relequestual:
Thank you for your contribution!

benjagm (approved):
Thanks a lot for adding this!

pages/understanding-json-schema/reference/combining.md
```diff
@@ -135,6 +135,8 @@ Multiple of *both* 5 and 3 is rejected.
 15
 ```
 
+Careful consideration should be taken when using `oneOf` entries as the nature of it requires verification of all sub-schemas so doing anything recursive can have a severe performance impact.
```
gregsdennis, line 138: As I mhad mentioned in the issue, the concern is not recursive schemas.  Yes, that can increase iterations, but the primary thing here is that `oneOf` _requires_ evaluating all of the schemas.  The preference is `anyOf` when possible.

I think that separately, recursive schemas should have a warning.

---

Also, I'd be interested in what recursive reference inside a `oneOf` is.  Typically, this will result in the kind of recursion that's disallowed by the spec, e.g. `{ "$ref": "#" }` as this produces infinite loops.

unguul: > I think that separately, recursive schemas should have a warning.

I agree. A general warning about recursive schemas is warranted, and a separate one for the use of `oneOf`.

> Also, I'd be interested in what recursive reference inside a oneOf led you to file the issue initially

It's been a while since I opened that initial issue (https://github.com/justinrainbow/json-schema/issues/640) but the TLDR is that it's easy to make a schema that takes a long time to process, and it's not immediately obvious where the issue is.

gregsdennis: Oh, okay.  Yeah that schema is a lot like the notorious FHIR schema, which has been jokingly used as a benchmark among us library maintainers.  A lot of implementations just choke on it (including my previous implementation).

It has around 50 definitions and a `oneOf` that references most (if not all) of them, then a lot of the definitions have nested references back to the root.  It's horrible.  People shouldn't do that.

If their data looks like that, I'd say they have a larger problem with data design, and JSON Schema isn't going to help the situation.  In these cases, when JSON Schema exacerbates a problem, it's better to find the root cause and fix that rather than throwing another tool at it.

gregsdennis:
I think these notes are fine, but I wonder if we should wait for the new "alert"-style formatted boxes that will be introduced with https://github.com/json-schema-org/website/pull/314.  This seems like a good use case for those.

benjagm:
@unguul can you please modify the code to use the new custom tag "warning"

Example of usage:

````
<Warning>
Careful consideration should be taken when using `oneOf` entries as the nature of it requires verification of *every* sub-schema which can lead to increased processing times. Prefer `anyOf` where possible.
</Warning>
````

File: pages/understanding-json-schema/reference/combining.md [hunk 2]
```diff
@@ -135,6 +135,9 @@ Multiple of *both* 5 and 3 is rejected.
 15
 ```
 
+Careful consideration should be taken when using `oneOf` entries as the nature of it requires verification of *every* sub-schema
+which can lead to increased processing times. Prefer `anyOf` where possible.
```
benjagm, line 139:
```suggestion
<Warning>
Careful consideration should be taken when using `oneOf` entries as the nature of it requires verification of *every* sub-schema which can lead to increased processing times. Prefer `anyOf` where possible.
</Warning>
```

benjagm (changes requested):
Can you please check my suggestion to use here the new custom tags for better styling?

pages/understanding-json-schema/reference/combining.md
benjagm: @unguul please proceed with this change and we are ready to go

benjagm (changes requested):
Just a small suggestion and we'll be ready yo merge!

pages/understanding-json-schema/reference/combining.md
gregsdennis: @unguul one change remaining, then this can be merged.

benjagm (approved):

unguul:
@benjagm I apologize for the silence. I was offline for a couple of weeks. Thanks for taking care of this :pray:

PR #342, opened 2024-02-23
ci: added triage labels for issues

### LINKED ISSUE #341: Make sure all the new issues are created with the label "Status: Triage"
#### Reason/Context
it will be great to make sure all the new issues are created with the label "Status: Triage". This can improve collaboration and help with the adoption of the new triage process. Ideally we'll use Github actions.

For reference:
https://docs.github.com/en/actions/managing-issues-and-pull-requests/adding-labels-to-issues

**Do you think this work might require an [Architectural Decision Record (ADR)](https://github.com/json-schema-org/community/blob/main/CONTRIBUTING.md#key-architectural-decisions)? (significant or noteworthy)**

No

utnim2:
GitHub Issue: #341 
Fixed #341 

**Summary**:
All the new issue are being created with label triage 

No
[JUSTIFICATION]


gregsdennis:
I've left a comment in the issue with some alternatives.

Relequestual:
Given the number of additional commits, please make sure this PR is "squashed and merged" when merging.

.github/issue-labler.yml
```diff
@@ -0,0 +1,3 @@
+Status: 
+  Triage:
+    - '(Describe the bug)'
```
gregsdennis, line 3: We probably want more here to cover enhancements, etc.  We can add those as we think of them, though.

gregsdennis (approved):

.github/workflows/issue-labeler.yml
```diff
@@ -0,0 +1,19 @@
+name: Issue Labeler
+on:
+  issues:
+    types: [opened, edited]
```
Relequestual, line 4: This should only be on opened.

.github/workflows/issue-labeler.yml
```diff
@@ -0,0 +1,19 @@
...
+  contents: read
+
+jobs:
+  triage:
+    runs-on: ubuntu-latest
+    steps:
+    - uses: github/issue-labeler@v3.1 
+      with:
+        configuration-path: .github/issue-labeler.yml
+        not-before: 2020-01-15T02:54:32Z
```
Relequestual, line 17: Having "not-before" doesn't make sense.
It looks like you copied and pasted this from the example workflow.

.github/workflows/issue-labeler.yml
```diff
@@ -0,0 +1,19 @@
...
+        enable-versioned-regex: 0
```
Relequestual, line 18: What does this do?

.github/workflows/issue-labeler.yml
```diff
@@ -0,0 +1,19 @@
...
+
+permissions:
+  issues: write
+  contents: read
+
+jobs:
+  triage:
+    runs-on: ubuntu-latest
+    steps:
+    - uses: github/issue-labeler@v3.1
```
Relequestual, line 14: Let's use the latest version =]

.github/issue-labler.yml
Relequestual, line 3: Where did you get this regex from?
We want all opened issues to gain the label.
It should be all opened issues, per https://github.com/github/issue-labeler?tab=readme-ov-file#label-all-issues

.github/issue-labler.yml
```diff
@@ -0,0 +1,3 @@
+Status: 
+  Triage:
```
Relequestual, line 2: Do the docs say you can do this to create a nested label name?
Remember, the label is not just `Triage`.

Relequestual (changes requested):
Thanks for your work on this so far. I think we have a way to go yet. I've made some suggestions and raised some questions.

.github/issue-labler.yml
Min2who: >  It should be all opened issues, per https://github.com/github/issue-labeler?tab=readme-ov-file#label-all-issues

You were right

.github/issue-labler.yml
Min2who: > Do the docs say you can do this to create a nested label name? Remember, the label is not just `Triage`.

what is the alternative then ?

.github/issue-labler.yml
princerajpoot20, line 3:
```suggestion

```
What is the use of this file? I don't think we need this file, as we want to add labels to all issues, not just to specific types.

.github/workflows/issue-labeler.yml
princerajpoot20, line 4:
```suggestion
    types: [opened]
```

.github/workflows/issue-labeler.yml
```diff
@@ -0,0 +1,19 @@
+name: Issue Labeler
+on:
+  issues:
+    types: [opened, edited]
+
+permissions:
+  issues: write
+  contents: read
```
princerajpoot20, line 8:
```suggestion
```
We don't need read permission, as we want to add labels to all issues irrespective of their content.

.github/workflows/issue-labeler.yml
princerajpoot20, line 18:
```suggestion

```
We don't really need all this. I would suggest using a much simpler approach for this. We can simply add label using `github-script`

princerajpoot20 (changes requested):
@Min2who few suggestions 🙂
cc: @benjagm @gregsdennis @Relequestual

.github/issue-labler.yml
gregsdennis: It's configuration.  See https://github.com/github/issue-labeler

princerajpoot20: @gregsdennis, do we need to add this label only for bugs? It will not add a label for the other issues. 
Also, I have tried testing this code in my repo, but it's not working. @Min2who, could you please share the workflow run link? Here is the link to [my run](https://github.com/princerajpoot20/CI-CD-Testing/actions/runs/8032729704/)

gregsdennis: See the heading "Label All Issues" in the linked docs.

.github/workflows/issue-labeler.yml
gregsdennis: Using the action sets up for additional label automation later if we want it.

.github/workflows/issue-labeler.yml
gregsdennis: It's fine to leave this.

.github/issue-labler.yml
princerajpoot20: @gregsdennis Sorry for this, my bad for not getting this. 
Just one thing to confirm. In the documentation of `all labels`, you have to write `/.*/`. If we write to Describe the bug, it will be the case of a simple example. Sorry if I am missing something.
Screenshot 2024-02-25 at 1 16 59 AM

gregsdennis: The requirement is to add the label to all issues.  The `/.*/` text is just a regex that matches everything.

The config specifies a label and a regex to match. If the regex matches the content of the issue, the label is applied.

So by specifying a regex that matches everything, the label is applied to all issues, which is the requirement.

.github/workflows/issue-labeler.yml
gregsdennis: We actually do need this because the action needs to read the issues.

.github/workflows/issue-labeler.yml
gregsdennis: I would say that we probably don't want to backfill-label every issue.  We should set this to whatever UTC is for this (PR open)
image

... or perhaps some date that we _do_ want to backfill with the label.

.github/workflows/issue-labeler.yml
gregsdennis: Best I can see is https://github.com/github/issue-labeler?tab=readme-ov-file#example-using-versioned-issue-templates, but I'm also not sure.  I don't think it applies to what we're doing, but I also don't know whether we have to explicitly state that it's turned off.

.github/workflows/issue-labeler.yml
gregsdennis: Looks like 3.4 is latest

.github/issue-labler.yml
gregsdennis: > nested label name

I'm not sure what this means, but the label, per the issue, should be "Status: Triage".  You'll need to use quotes because `:` is a special char in YAML.

```suggestion
'Status: Triage':
```

Min2who:
I have pushed the changes above and this is my [test run](https://github.com/Min2who/CI-CD-test/actions/runs/8036688999)

cc 
@gregsdennis @Relequestual @benjagm @princerajpoot20

benjagm:
> I have pushed the changes above and this is my [test run](https://github.com/Min2who/CI-CD-test/actions/runs/8036688999)

Great progress @Min2who and @princerajpoot20.  This seems to work. Lats thing to check: Does this action require to configure a secret to store the token?

gregsdennis:
@benjagm no, the token is configured by Github [automatically](https://docs.github.com/en/actions/security-guides/automatic-token-authentication) at build start.

gregsdennis (approved):

gregsdennis:
@benjagm @Relequestual please refresh your reviews.

benjagm (approved):
Good job!

gregsdennis:
@Relequestual I'm clearing your review as everything has been addressed.  @benjagm and I have approved.

Relequestual (changes requested):
Looks almost all the way there. Just one small suggestion.

.github/workflows/issue-labeler.yml
```diff
@@ -0,0 +1,19 @@
...
+
+jobs:
+  triage:
+    runs-on: ubuntu-latest
+    steps:
+    - uses: github/issue-labeler@v3.4 
+      with:
+        configuration-path: .github/issue-labeler.yml
+        not-before: 2024-02-23T23:24:00+13:00
```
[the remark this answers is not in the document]
utnim2: done @Relequestual 👍

gregsdennis:
@Relequestual this needs a refreshed review.

Relequestual (approved):
Thanks for your work on this!

Issue #349, opened 2024-02-24
[docs]: Inconsistent code rendering in documentation examples


DarhkVoyd:
#### Describe the bug

While navigating through the getting started docs, I observed a formatting inconsistency in the examples. When the code snippet is the entire piece of code, it is properly rendered with colors and indentations. However, in cases where the code is only a part of the entire example and is denoted by "..." to indicate existing code, it appears as plain text without proper formatting.

#### How to Reproduce

1. Navigate to the getting started documentation.
2. Look for examples using "..." to denote existing code snippets.
3. Notice that these snippets are rendered as plain text without colors or indentations.

#### Screenshot
![Screenshot 2024-02-23 at 10 33 46 PM](https://github.com/json-schema-org/website/assets/77478658/57ae4e12-125a-45ce-b6d6-1bfa82fe1a6e)

#### Expected behavior

I expect that all code snippets, whether part of the entire example or not, should be consistently rendered with proper code formatting, including colors and indentations.

Sahil-Shadwal:
Please assign me this issue , Thanks.

DarhkVoyd:
Hey @Sahil-Shadwal 
This issue is currently is in `triage`. Kindly wait till a maintainer marks it `available`.

benjagm:
Expectation with this issue is format the json code properly if it is not formatted and if possible show "json" instead of "data".

benjagm:
The issue is ready. @DarhkVoyd are you ok with @Sahil-Shadwal fixing this and you be the reviewer?

DarhkVoyd:
@benjagm Yeah, sure!

benjagm:
All yours @Sahil-Shadwal! Please go ahead!

benjagm:
@Sahil-Shadwal are you still working on it?

Adity20:
Okay @benjagm

Sahil-Shadwal:
hey @benjagm, i am in mid of my exams right now . So if you want you can assign this issue to someone else. I hope this won't leave a bad impression . Will be looking forward to contribute in another issue.

DhairyaMajmudar:
If it's okay shall I continue this one : )

benjagm:
@Sahil-Shadwal not a bad impression at all!! We hope you exams go well. That is the priority right now. Good luck!

@DhairyaMajmudar will fix this one. And as soon as you are done with the exams just let me know if you can find issues to complete your qualification task.

benjagm:
[...]

PR #356, opened 2024-02-25
Fix Inconsistent code rendering in documentation examples

Sahil-Shadwal:
**GitHub Issue:** #349 

**Summary**:
The documentation inconsistent code rendering examples have been fixed.

No

### DIFF
--- a/pages/learn/getting-started-step-by-step.md
+++ b/pages/learn/getting-started-step-by-step.md
@@ -139,6 +139,7 @@ With the new `properties` validation keyword, the overall schema looks like this
   "description": "A product from Acme's catalog",
   "type": "object",
   "properties": {
+    ...
     "productId": {
       "description": "The unique identifier for a product",
       "type": "integer"
@@ -157,6 +158,7 @@ The following example adds another required key, `productName`. This value is a 
   "description": "A product from Acme's catalog",
   "type": "object",
   "properties": {
+    ...
     "productId": {
       "description": "The unique identifier for a product",
       "type": "integer"
@@ -195,6 +197,7 @@ To define a required property:
 
 ```json
 {
+  ...
   "price": {
     "description": "The price of the product",
     "type": "number",
@@ -207,6 +210,7 @@ To define a required property:
 
 ```json
 {
+  ...
   "price": {
     "description": "The price of the product",
     "type": "number",
@@ -228,6 +232,7 @@ With the new `required` keyword and `price` key, the overall schema looks like t
   "description": "A product from Acme's catalog",
   "type": "object",
   "properties": {
+    ...
     "productId": {
       "description": "The unique identifier for a product",
       "type": "integer"
@@ -277,6 +282,7 @@ To define an optional property:
 
 ```json
 {
+  ...
   "tags": {
     "description": "Tags for the product",
     "type": "array",
@@ -291,6 +297,7 @@ To define an optional property:
 
 ```json
 {
+  ...
   "tags": {
     "description": "Tags for the product",
     "type": "array",
@@ -306,6 +313,7 @@ To define an optional property:
 
 ```json
 {
+  ...
   "tags": {
     "description": "Tags for the product",
     "type": "array",
@@ -328,6 +336,7 @@ With the new `tags` keyword, the overall schema looks like this:
   "description": "A product from Acme's catalog",
   "type": "object",
   "properties": {
+    ...
     "productId": {
       "description": "The unique identifier for a product",
       "type": "integer"
@@ -394,6 +403,7 @@ To create a nested data structure:
   "dimensions": {
     "type": "object",
     "properties": {
+      ...
       "length": {
         "type": "number"
       },
@@ -415,6 +425,7 @@ To create a nested data structure:
   "dimensions": {
     "type": "object",
     "properties": {
+      ...
       "length": {
         "type": "number"
       },
@@ -440,6 +451,7 @@ Using the new nested data structures, the overall schema looks like this:
   "description": "A product from Acme's catalog",
   "type": "object",
   "properties": {
+    ...
     "productId": {
       "description": "The unique identifier for a product",
       "type": "integer"
@@ -501,6 +513,7 @@ The following schema validates a geographical location:
   "required": ["latitude", "longitude"],
   "type": "object",
   "properties": {
+    ...
     "latitude": {
       "type": "number",
       "minimum": -90,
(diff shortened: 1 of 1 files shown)


DarhkVoyd:
Thanks for this. I'll review the changes soon. Meanwhile, please link this PR to the original issue.
Reference: https://docs.github.com/en/issues/tracking-your-work-with-issues/linking-a-pull-request-to-an-issue

DarhkVoyd (changes requested):
Hey, Thank you for this one. Kindly, review on my comments.
1. I believe your editor's markdown linter has changed certain parts of the markdown. Namely, changing `*` to `-` and changing floating point numbers such as `12.50` to `12.5` and some of the indentation. They should not be a part of the PR, kindly revert those.
2. Your approach is mostly valid, but the goal of previous syntax was to demonstrate that the json key is being built upon the previous step's json key or the schema object itself so that it is clearly visible that those are relative and nested changes. This is no longer explicit. We could use some more `...` to make it explicit. 
3. Lastly, if possible show "json" instead of "data".

DarhkVoyd:
@benjagm Since, the solution involves closing the json, I think we still need to add some `...` so that its abundantly clear that we are building on top of certain sections of schema. Also, I am waiting for an update about `data` to `json`.

Sahil-Shadwal:
Okay will revert those .

DarhkVoyd:
@Sahil-Shadwal Let's wait for the maintainers opinion on my suggestions.

benjagm:
Thanks everyone!!

Why changing the format of the list to use '-' instead '*'? That has nothing to do with the issue and doesn't fix anything.

Regarding the '...' that is true @DarhkVoyd . We need to keep them there.

Sahil-Shadwal:
Sorry about * to - it got auto changed when i saved the file , let me correct that . Is there anything else?

DarhkVoyd:
@Sahil-Shadwal Kindly refer to my previous review. Regarding `...`, please add `...` at appropriate places to ensure that it's abundantly clear that we are building on top of certain sections of schema and clearly express its structure. 

> Hey, Thank you for this one. Kindly, review on my comments.
> 
> 1. I believe your editor's markdown linter has changed certain parts of the markdown. Namely, changing `*` to `-` and changing floating point numbers such as `12.50` to `12.5` and some of the indentation. They should not be a part of the PR, kindly revert those.
> 2. Your approach is mostly valid, but the goal of previous syntax was to demonstrate that the json key is being built upon the previous step's json key or the schema object itself so that it is clearly visible that those are relative and nested changes. This is no longer explicit. We could use some more `...` to make it explicit.
> 3. Lastly, if possible show "json" instead of "data".

pages/learn/getting-started-step-by-step.md
```diff
@@ -99,28 +99,32 @@ Using the product catalog example, `productId` is a numeric value that uniquely
 To add the `properties` object to the schema:
 
 1. Add the `properties` validation keyword to the end of the schema:
-    
+
 ```json
-...
-"title": "Product",
```
gregsdennis, line 105: If we're going to make these syntax-highlighted, we need to do this differently.

    ```jsonc
    // ...
    "title": "Product",

`jsonc` allows comments, and I'm pretty sure we can get it to have the same highlight colors as `json` (probably in another PR if it's not already that way)

pages/learn/getting-started-step-by-step.md
```diff
@@ -490,21 +531,25 @@ The following schema validates a geographical location:
...
-"properties": {
-...
-  "warehouseLocation": {
+{
+  "properties": {
+  ...
+    "warehouseLocation": {
+    }
   }
 }
```
gregsdennis, line 542: Similarly, this isn't the full schema on its own.  The current snippet (although not proper JSON) demonstrates well that this is an excerpt from the above schema, but the new snippet is a schema by itself.

This is seen in a lot of these changes.

pages/learn/getting-started-step-by-step.md
```diff
@@ -346,59 +375,69 @@ The earlier examples describe a flat schema with only one level. This section de
...
 ```json
-"dimensions": {
-  "type": "object",
-  "properties": {
-    "length": {
-      "type": "number"
-    },
-    "width": {
-      "type": "number"
+{
```
gregsdennis, line 424: like here

pages/learn/getting-started-step-by-step.md
```diff
@@ -346,59 +375,69 @@ The earlier examples describe a flat schema with only one level. This section de
...
-    "length": {
-      "type": "number"
-    },
-    "width": {
-      "type": "number"
-    },
-    "height": {
-      "type": "number"
+{
+  "dimensions": {
```
gregsdennis, line 403: here

pages/learn/getting-started-step-by-step.md
```diff
@@ -346,59 +375,69 @@ The earlier examples describe a flat schema with only one level. This section de
...
 ```
 
 2. Define the `type` validation keyword as `object`:
-    
+
 ```json
-"dimensions": {
-  "type": "object",
+{
+  "dimensions": {
```
gregsdennis, line 393: and here

gregsdennis (changes requested):
I think we need to re-evaluate the importance of having proper JSON.  I'm not sure it's needed.  More important than the snippets being "correct" is that they are understandable.  I think many of these edits serve to confuse the reader rather than clarifying the ideas that the narrative is trying to convey.

Sahil-Shadwal:
Thanks for the review, will surely look into it . But it is getting confusing, should i revert back  the changes ?

DarhkVoyd:
@gregsdennis Hey, I am bit confused. As per my understanding, we agree on:

> The current snippet (although not proper JSON) demonstrates well that this is an excerpt from the above schema, but the new snippet is a schema by itself.

So are we against closing the json as a solution, or the usage of `...` or the entire approach.

gregsdennis:
I'm happy to have the `...`, but the changes look like they're trying to create full JSON values, which isn't right.

It looks like the code blocks already have the language specifier.  I'm guessing the bare `...` is causing them to be rendered as plain text.

I think if we change the language specifier to `jsonc` and use `// ...` it _might_ render right.

Otherwise we'll have to do something like

    ```jsonc
    {
      // ...
      "properties": {
        // ...
        "newProperty": {}
      }
    }
    ```

Then we need to make sure the `jsonc` formatting is the same as the `json` formatting.

DarhkVoyd:
Yes, that's what I meant. You explained it better.

benjagm:
@Sahil-Shadwal is this ready to review? Thanks a lot!

gregsdennis:
@Sahil-Shadwal was this replaced by another PR? Could you link to it, please?

DarhkVoyd:
This was replaced with this PR:  https://github.com/json-schema-org/website/pull/475

Issue #282, opened 2024-01-25
Add a new section in the Docs for other JSON Schema resources


benjagm:
#### Description
If would be great to have resources section to show articles, papers, books podcasts and any other type of resources. We'll use json to store the data and provide dynamic UX.

Content types: 
* Articles
* Books
* Courses
* Videos
* Podcasts
* Papers

Data structure example:
````
- title: Using JSON Schema
  url: https://books.apple.com/us/book/using-json-schema/id903248630
  year: 2014
  type: books
  summary: Learn and Apply JSON Schema by Example, with JavaScript (Node.js) and Python Programs
- title: JSON at Work
  url: https://www.oreilly.com/library/view/json-at-work/9781491982389/
  year: 2017
  type: books
  summary: A comprehensive overview of the JSON ecosystem, including JSON Schema
````
An amazing source of information will be: https://github.com/sourcemeta/awesome-jsonschema/blob/master/data.yaml

A possible prototype for this may be:

Resources menu:
<img width="356" alt="Screenshot 2024-02-10 at 18 51 04" src="https://github.com/json-schema-org/website/assets/40007659/db480fbd-5e27-4506-b989-2e64b256fb13">

Resource category page (example):
![Untitled presentation (1)](https://github.com/json-schema-org/website/assets/40007659/d562292c-5a63-4e4e-a8cd-3983cfcfe49a)

**Do you think this work might require an [Architectural Decision Record (ADR)](https://github.com/json-schema-org/community/blob/main/CONTRIBUTING.md#key-architectural-decisions)? (significant or noteworthy)**

No

benjagm:
I have some progress in this draft PR #281 but I'd love to have someone to continue the development.

Min2who:
> I have some progress in this draft PR #281 but I'd love to have someone to continue the development.

@benjagm can i help you with it?

benjagm:
> can i help you with it?

@Min2who of course! Thanks a lot!! Please go ahead. (check my draft PR with the side menu logic)

Akshaybagai52:
Hey @benjagm 
Can I work on this issue
Thanks

benjagm:
> @benjagm can I look into this?

Available to work on this one?

Atharva-Kanherkar:
Hi, if no one is working on this, i would like to work on this. 
Seems like a good issue to get hands on with the codebase!
> @Akshaybagai52 Available to work on this one?

benjagm:
[...]

PR #395, opened 2024-02-29
feat: added resource section with article and books routes working

Akshaybagai52:
**GitHub Issue:** #282

**Summary**:

Yes/No
[JUSTIFICATION]

### DIFF
--- a/components/Sidebar.tsx
+++ b/components/Sidebar.tsx
@@ -5,6 +5,7 @@ import Link from 'next/link';
 import { HOST } from '~/lib/config';
 import classnames from 'classnames';
 import { SegmentHeadline } from './Layout';
+import extractPathWithoutFragment from '~/lib/extractPathWithoutFragment';
 import CarbonAds from './CarbonsAds';
 
 const DocLink = ({
@@ -19,7 +20,7 @@ const DocLink = ({
   url.search = '';
   url.hash = '';
   const stringUrl = url.toString().substr(HOST.length, Infinity);
-  const isActive = uri === stringUrl;
+  const isActive = uri === extractPathWithoutFragment(stringUrl);
   return (
     <Link
       href={uri}
@@ -45,7 +46,7 @@ const DocLinkBlank = ({
   url.search = '';
   url.hash = '';
   const stringUrl = url.toString().substr(HOST.length, Infinity);
-  const isActive = uri === stringUrl;
+  const isActive = uri === extractPathWithoutFragment(stringUrl);
   return (
     <Link
       href={uri}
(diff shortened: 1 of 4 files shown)


.gitmodules
```diff
@@ -40,3 +40,6 @@
 [submodule "_includes/dot-github"]
 	path = _includes/dot-github
 	url = https://github.com/json-schema-org/.github.git
+[submodule "path/to/submodule"]
```
benjagm, line 43: We should not import this data as submodule. We like to avoid this type of dependency from projects outside our org. We'll do our own version of the data file in our data folder.

components/Sidebar.tsx
```diff
@@ -312,10 +324,32 @@ export const DocsNav = () => {
...
+            <SegmentHeadline label='Resources' />
+          </div>
+          <svg id='arrow' className='arrow' style={{ transform: rotateResources, transition: 'all 0.2s linear' }} xmlns='http://www.w3.org/2000/svg' fill='none' height='32' viewBox='0 0 24 24' width='24'><path clipRule='evenodd' d='m16.5303 8.96967c.2929.29289.2929.76777 0 1.06063l-4 4c-.2929.2929-.7677.2929-1.0606 0l-4.00003-4c-.29289-.29286-.29289-.76774 0-1.06063s.76777-.29289 1.06066 0l3.46967 3.46963 3.4697-3.46963c.2929-.29289.7677-.29289 1.0606 0z' fill='#707070' fillRule='evenodd' /></svg>
+
+        </div>
+        <div
+          className={classnames( 'ml-6', { 'hidden': !active.getResources })}
+          id='resources'
+        >
+          <DocLink uri='/resources/books' label='Books' />
```
benjagm, line 341: These are static links. I'd like to avoid this. Instead I'd like you to go through the data and extract the resource categories available there to build dynamically these links.

pages/resources/articles/index.page.tsx
```diff
@@ -0,0 +1,80 @@
+import React, { useEffect, useState } from "react";
```
benjagm, line 1: I expect one single generic resources page that accepts the resource type as parameters os it will work for all cases.

pages/resources/books/index.page.tsx
```diff
@@ -0,0 +1,76 @@
+import React, { useEffect, useState } from "react";
```
benjagm, line 1: I expect one single generic resources page that accepts the resource type as parameters os it will work for all cases.

benjagm (changes requested):
Please address this first wave of changes and we'll continue the review of UX when this is done.

PR #494, opened 2024-03-09
fixed the footer space issue

Ali7040:
remove the mx-auto form sidebar because we have mx-4 and mx-auto both in same line. and fix the footer space issue.

**Issue Number:**

-  Closes ##455 
-  Related to #___ 
-  Others? 

**Screenshots/videos:**

**Summary**

### DIFF
--- a/components/Layout.tsx
+++ b/components/Layout.tsx
@@ -61,7 +61,7 @@ export default function Layout({
         <main
           className={classnames(
             mainClassName,
-            'z-10 h-[100vh] xl:rounded-xl pt-4 mx-auto',
+            'z-10 h-screen  xl:rounded-xl pt-4 mx-auto',
           )}
         >
           <header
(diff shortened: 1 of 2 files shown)


DhairyaMajmudar (approved):
Here in place of `h-[100vh]` you can use  `h-screen` both will work same but the suggested one is a good practice to use.

Rest looks good to me  : )

Ali7040:
> Here in place of `h-[100vh]` you can use `h-screen` both will work same but the suggested one is a good practice to use.
> 
> Rest looks good to me : )
Thanks for your feedback. I will update it.

Ali7040:
> Here in place of `h-[100vh]` you can use `h-screen` both will work the same but the suggested one is a good practice to use.
> 
> Rest looks good to me : )

Done :)

benjagm (approved):
Good job!

officeneerajsaini:
Great @Ali7040

Issue #440, opened 2024-03-04
✨ Enhancement: Page in overview section to show Case Studies


benjagm:
### Is your feature request related to a problem? Please describe

No

### Describe the solution you'd like

it would be amazing to have a new page inside the overview section to show the case studies published in the blog.
This will be a data drive page reading results from a json file stored in our data folder. Each item of the file will have these fields:
- logo
- summary
- List of links
  - lang : Language in ISO 639-1
  - url : link to the case study in that language
  
Please see this prototype:
<img width="542" alt="Screenshot 2024-03-04 at 15 45 26" src="https://github.com/json-schema-org/website/assets/40007659/ca814769-a895-46a3-8558-3893172e5601">

This is a prototype made by screenshot. Please complete the job using styles as per our site.

For the cards please consider using border shadows similar to the ones in  the landing page:
<img width="1019" alt="Screenshot 2024-03-04 at 15 50 28" src="https://github.com/json-schema-org/website/assets/40007659/dfde3d13-750c-4424-b3e9-c0d60aa50c51">

The case studies are available in the blog:
https://json-schema.org/blog?type=Case%20Study

### Are you working on this?

No

TamannaVerma99:
Hey!
Since my previous issue is blocked by another one (the card component is yet to be made) , Is it ok if I can be assigned this issue?

benjagm:
Please go ahead. These cards is not going to be the standard ones. So no blockers expected here.

TamannaVerma99:
@benjagm, Had a few questions related to this one.

In order to add a custom card component which will read data from the specified JSON file, I would either have to use a .tsx file or a .mdx file. However, the present configuration for the pages/SideBar section specifies the use of only '.md' files.  I tried finding out ways to implement components in .md itself, but conversion to .mdx is mandatory and based on the settings in [slug].page.tsx file (specifically, `getStaticMarkdownPaths` property) allows only .md files to be rendered. 

So, will it be okay to modify this function to include .mdx files or is there another way round that can be used?

benjagm:
> I would either have to use a .tsx

Please see the [slug].page.tsx in each subfolder of pages. This can help you think about how you can have your own page without md. I will recommend you to take a look to [implementers page](https://github.com/json-schema-org/website/blob/main/pages/implementations/index.page.tsx) but be careful, this page doesn't have sidebar.

I hope this helps. We have multiple issues implementing similar features. Pleas see #429

TamannaVerma99:
[...]

PR #540, opened 2024-03-14
Added Case Studies page in Overview Section 

TamannaVerma99:
Added a new page in Overview section to show the case studies published in the blog. Currently, since I don't have access to the data file that has information related to blogs published, I have created a sample file of my own named `casestudies.json` with the same data format as mentioned in the issue and linked it to the card component. This makes the JSON file easily replaceable with the type of data we'll be using.

**Issue Number:**

-  Related to #440  

No


benjagm:
@TamannaVerma99 Hi reviewed the PR and decided to push some changes to accelerate the process. The only pending thing on your side is to adjust the generic card component to make sure the logos shows properly. I like the logos much bigger with the right proportion bacause some of the logos show deformed.

What I fixed:

- Modified the json data to remove posts unrelated to case studies and updated the logos.
- Added all the logos.
- Modified how you was using the card component to use the image property to show the logo, instead of the icon as you was using. Now the behaviour is correct.

You did an amazing job and we are very very close.
Congrats!!

benjagm (changes requested):
Nice job. We are very very close.

See my review comments [here](https://github.com/json-schema-org/website/pull/540#issuecomment-2001921707)

TamannaVerma99:
> Nice job. We are very very close.
> 
> See my review comments [here](https://github.com/json-schema-org/website/pull/540#issuecomment-2001921707)

Thanks :) I will do the required changes asap.

components/Cardforcasestudies.tsx
```diff
@@ -0,0 +1,55 @@
...
+      <div className='flex justify-center '>
+        {image && <img src={image} className='h-32 w-36 p-2' />}
+      </div>
+      <div className='flex flex-row items-start mb-6'>
+        {icon && (
+          <span className='mr-6 flex h-14 w-14 flex-shrink-0 items-center justify-center rounded-lg border bg-blue-200 px-3 text-gray-900'>
+            <img src={icon} alt={title} className='h-full w-full' />
+          </span>
+        )}
+        <p className='mt-1 items-center text-[0.9rem] font-bold text-gray-900'>
```
Michael-Obele, line 24: Hey there! Great work on the code! I noticed the header seemed a bit small. Might we consider text-h5 or text-h6 for better readability?
```suggestion
        <p className='mt-1 items-center text-h5 font-bold text-gray-900'>
```

TamannaVerma99: Hey! Thanks for the feedback .I will work on it. Thanks:)

Michael-Obele:
Hi everyone,
I realize now that the new card component was likely created to address the issue of long titles exceeding the character limit. This is something I hadn't considered initially.

I'd be happy to adjust the existing card component by adding a conditional statement. This would check the word count and automatically adjust the text size to be larger for shorter content and smaller for longer content.

To keep things moving smoothly, I can either open a separate issue for this or jump right in and make the fix here in the PR.  Just let me know which approach you prefer!

VivekJaiswal18:
> Hi everyone, I realize now that the new card component was likely created to address the issue of long titles exceeding the character limit. This is something I hadn't considered initially.
> 
> I'd be happy to adjust the existing card component by adding a conditional statement. This would check the word count and automatically adjust the text size to be larger for shorter content and smaller for longer content.
> 
> To keep things moving smoothly, I can either open a separate issue for this or jump right in and make the fix here in the PR. Just let me know which approach you prefer!

But if we keep the title text size variable will it not make the UI inconsistent, if in the same page we have heading with variable length

Michael-Obele:
> > Hi everyone, I realize now that the new card component was likely created to address the issue of long titles exceeding the character limit. This is something I hadn't considered initially.
> > I'd be happy to adjust the existing card component by adding a conditional statement. This would check the word count and automatically adjust the text size to be larger for shorter content and smaller for longer content.
> > To keep things moving smoothly, I can either open a separate issue for this or jump right in and make the fix here in the PR. Just let me know which approach you prefer!
> 
> But if we keep the title text size variable will it not make the UI inconsistent, if in the same page we have heading with variable length

You raise a good point about UI consistency.  While increasing the base font size slightly could help with longer titles, perhaps a maximum character limit or using ellipsis for overflow could be additional solutions to explore.

Michael-Obele:
Although I don't see any issue with having a separate card component as we have now, this is only a suggestion.

TamannaVerma99:
> Although I don't see any issue with having a separate card component as we have now, this is only a suggestion.

Hey! Actually I have also been working on welcome page ,for there too I have same issues if my title is even "Getting Started" still its covering two lines so it would be really great if we find any possible solution for it .I think we can also pass title heading by props.

benjagm:
This code has been already merged in the dev branch. Awesome work!

components/Card.tsx
```diff
@@ -0,0 +1,55 @@
...
+  title: string;
+  body: string;
+  icon?: string;
+  link?: string;
+  image?: string;
+}
+
+const CardBody = ({ title, body, icon, link, image }: CardProps) => {
+  return (
+    <div className='group relative h-full w-full max-w-lg rounded-lg border border-gray-200 bg-white p-6 px-12 shadow-3xl transition-colors delay-[150ms] ease-in-out hover:bg-slate-100'>
```
VivekJaiswal18, line 14: @TamannaVerma99 Please add `dark:text-black` in the div. As in dark mode the text is not visible.
image

PR #547, opened 2024-03-15
Fix broken links to header anchors

kurtmckee:

This PR fixes two broken relative links:

* The link definitions used hyphenated anchors (like `"additional-properties"`)
* The target headers define non-hyphenated IDs (like `"additionalproperties"`)

You can see the current problem by visiting:

https://json-schema.org/understanding-json-schema/reference/object#properties

and attempting to click on the "Additional Properties" and "Unevaluated Properties" link in the call-out text.

image

`git blame` showed that the hyphenated anchor lines were modified a few months ago, while the target lines were last modified ~2 years ago, so I removed the hyphens to align with the long-standing target lines.


benjagm (approved):
Thanks a lot for finding and fixing this!! ❤️

PR #612, opened 2024-03-27
Remove useless page

benjagm:
We are removing /learn/test page because the content is not adding any value.


vinfinity7:
Hi @benjagm  , any more refinements you think of implementing now ?

PR #493, opened 2024-03-09
[Feat]: Add Community Page

officeneerajsaini:
I make a new saprate community page

**What kind of change does this PR introduce?**
Feature

**Issue Number:**

-  Closes #280

PR #633, opened 2024-04-08
[Feat]: Add Community Page #493

### LINKED ISSUE #280: Create a new page for Community inside the Overview Section of the Docs.
#### Description
Create a new page for Community inside the Overview Section of the Docs.
 * Explaining all community channels.
 * All community meetings.
 * Blog.
 * All community programs.
 * Members of the tsc, contributors and ambassadors.

**Do you think this work might require an [Architectural Decision Record (ADR)](https://github.com/json-schema-org/community/blob/main/CONTRIBUTING.md#key-architectural-decisions)? (significant or noteworthy)**

No

officeneerajsaini:
I make a new separate community page

Feature

**Issue Number:**

-  Closes #280


benjagm (changes requested):
After seeing the lot of unnecessary changes coming IT think will be better to create a fork from web-release-3 and  manually add the community folder inside pages
otherwise merging this will be a nightmare.

In addition, I saw you adding multiple jsx files inside community which I think is not ideal for 2 reasons:

1. Let's use typescript instead javascript

2. Why breaking down into components that are not reusable at all? If those components are flexible and reusable maybe they can go into components but I seems that they are not reusable at all so better to avoid that breakdown of pieces to put all the logic inside /community/index.page.tsx

Issue #266, opened 2022-07-13
Create a JSON Schema Glossary Page or Site


Julian:
It'd be nice if we had a glossary of terms -- a page, in not very technical jargon, which explains common terminology used by us in conversation, and which could be linked to by other pages or people in order to answer questions like "what is a vocabulary" in a succinct way.

Here are two examples of glossaries I've contributed to or used heavily which may be decent examples of the concept:

* [The Python glossary](https://docs.python.org/3/glossary.html)
* [The Lean glossary](https://leanprover-community.github.io/glossary.html#widget)

EDIT: The page has been published at https://json-schema.org/learn/glossary.html

To help, just grab any of the entries below which don't already exist on the page, and share that you're working on them!

---

List of Terms to Define

- [x] dialect json-schema-org/json-schema-org.github.io#484
- [x] instance json-schema-org/json-schema-org.github.io#462
- [x] vocabulary json-schema-org/json-schema-org.github.io#484
- [x] schema json-schema-org/json-schema-org.github.io#462
- [x] meta-schema json-schema-org/json-schema-org.github.io#469
- [x] keyword json-schema-org/json-schema-org.github.io#469
- [ ] hyper-schema
- [x] draft json-schema-org/json-schema-org.github.io#461
- [ ] ref
- [ ] annotation
- [ ] validation
- [ ] validation result
- [ ] annotation (keyword)
- [ ] validation (keyword)
- [ ] assertion (keyword)
- [ ] in-place application
- [ ] child application
- [x] subschema json-schema-org/json-schema-org.github.io#520
- [ ] resource / schema resource
- [ ] compound schema resource
- [ ] anchor
- [x] JSON json-schema-org/json-schema-org.github.io#521
- [ ] JSON pointer
- [ ] relative JSON pointer

bhavukkalra:
In the process of creating a drafts for `annotation` and `validation`. Will send over for review once ready

vickywane:
@benjagm I'd love to take up `anchor` and `ref` and submit a PR for them.

benjagm:
> @benjagm I'd love to take up `anchor` and `ref` and submit a PR for them.

Thanks a lot @vickywane ! Please go ahead

bhavukkalra:
Just created PR #668. To add the sections of `annotation` and `annotation (keyword)`. I tried to these two attributes in a single PR as these are closely related and one references the other.

Temidayo32:
[...]

PR #668, opened 2024-04-20
Add sections of annotation and annotation (keyword) in the glossary

bhavukkalra:
As discussed in the issue #266. This PR adds two very similar sections to the [glossary](https://json-schema.org/learn/glossary) section. 

**Screenshots/videos:**
image

**Summary**
This PR is a part of the individual PR's getting created for completing the `glossary` page. Used these as a reference to draft the description of these two related sections

- https://json-schema.org/understanding-json-schema/reference/annotations
- https://json-schema.org/draft/2020-12/json-schema-core#name-annotations

No

### DIFF
--- a/pages/learn/glossary.md
+++ b/pages/learn/glossary.md
@@ -149,6 +149,23 @@ The JSON Schema specification includes a number of vocabularies which cover each
 
 In some [dialects](#dialect) of JSON Schema, the `$vocabulary` keyword can be used to include the keywords defined by a vocabulary into the dialect, as well as to indicate whether implementations must specifically recognize the vocabulary in order to be able to process schemas written in the dialect or not.
 
+### Annotation
+
+Annotation is a piece of metadata attached to a piece of code, used to convey additional information about the code like its workings, the algorithm that it internally uses, etc. 
+
+
+Annotations provide supplementary information for developers to express intended usage, constraints, configurations or manage complexity, especially when the JSON Schema's definition gets enormous. 
+
+Different [keywords](#annotation-(keywords)) are defined within JSON Schema to convey various aspects of the schema, making it self-reliant and self-documenting.
+
+### Annotation (Keywords)
+
+Specific "Keys" acting as self-serving documentation for the defined schema by explaining the intended usage or constraints. 
+
+Annotation keywords can be used in any schema or sub-schemas as they don't affect any logical operations for the schema, such as validation, As they are only for the schema's "self-documenting" purposes. 
+
+Using these keywords is optional but recommended, as this makes your schema self-reliant. [Specific annotation keywords](../understanding-json-schema/reference/annotations) might depend on which draft version of the JSON schema you are employing.
+
 #### See also
 
 * [`json-schema-vocabularies`](https://github.com/json-schema-org/json-schema-vocabularies), a repository which collects known third-party JSON Schema vocabularies


gregsdennis (changes requested):
I'm not sure where the language of "self-reliant" and "self-documenting" is coming from.  (I'm not sure how "self-reliant" applies.)  Annotations can do more than just documentation.  They can provide context to validations ("Which subschema validated this data?"), they can append additional information to data (which is kinda what hyperschema does), and they can provide the host application with additional details for it to know how to continue processing the data. 

These definitions, to me, don't seem to align with what annotations are. 

Annotation keywords are merely keywords which produce annotations.

jdesrosiers:
I generally agree with Greg's review. I don't know what "self-reliant" means in this context. People often use annotation keywords (not annotations) to make a schema "self-documenting", but they're not really using them as annotations in that case. I would describe an annotation as meta-data associated with a particular value in a JSON instance. An annotation keyword is an instruction to the validator on how and where to associate an annotation to the instance. That annotation can be used for all sorts of things. It could even be used for different things depending on the context. So, people using `title` and `description` to document their schema ("self-documenting"), they may be using annotation keywords, but they aren't using annotations because the annotation doesn't exist until you apply the schema to an instance.

I'll give a simple example and a more complex one. Here's the simple example.

Given an instance:

```json
{
  "foo": "some value"
}
```

And a schema:

```json
{
  "type": "object",
  "properties": {
    "foo": {
      "type": "string",
      "title": "Foo"
    }
  }
}
```

If I apply the schema to the instance, the result is an annotated instance. I'm using a made-up syntax for annotated JSON. Hopefully it makes sense.

```javascript
{
  "foo": "some value" @title("Foo")
}
```

The value of `/foo` is `"some value"`, but that value now has a `title` annotation attached to it. In theory, you would pass this annotated instance to some process to do something with it that does something with that annotation.

For the more complex example, consider that annotations aren't always just the value of the annotation keyword, they could be computed in some way to get the actual value that will be attached to the instance. Here's a JSON Hyper-Schema example of attaching a link to an instance.

Start by retrieving an instance from an API: `https://example.com/api/widgets?page=1`

```http
HTTP/1.1 200 OK
Content-Type: application/json
Link: </schemas/widget-list>; rel="describedby"

{
  "list": [...],
  "page": 1,
  "per-page": 10,
  "next-page": 2
}
```

The instance identifies the schema that describes it using the "describedby" link, so we need to retrieve the schema as well.

```http
HTTP/1.1 200 OK
Content-Type: application/schema+json

{
  "type": "object",
  "properties": {
    "list": { "type": "array" },
    "page": { "type": "integer" },
    "per-page": { "type": "integer" },
    "next-page": { "type": "integer" }
  },
  "links": [
    { "rel": "next", "href": "/api/widgets?page={next-page}" }
  ]
}
```

Then we apply the schema to the instance to get an annotated instance:
```javascript
{
  "list": [...],
  "next-page": 2
} @links({ next: "https://example.com/widgets?page=2" })
```

Notice that the resulting meta-data fills in the URI Template and resolves the relative URI against the retrieval URI of the instance. A hyper-schema client can then use that annotated instance to get the next page in the list of widgets using the computed "next" link annotation.

```javascript
// Get an annotated instance
const widgetList = HyperSchema.get("https://example.com/api/widgets");

// Iterate over the list automatically following "next" links as necessary to get all the widgets.
for await (const widget of HyperSchema.iter(widgetList)) {
  console.log(widget);
}
```

This is is a fictitious library that I still delude myself that I'll find time to write one day, but it hopefully gives a more concrete example of the kinds of things a hyper-schema client could do with link annotations.

benjagm:
We did a good progress with the review of this PR. Are you still working on it Bhavuk?

benjagm:
We are closing this issue due to inactivity. Thanks everyone who was part of the discussions.

PR #671, opened 2024-04-22
This request is in response to issue #658.

### LINKED ISSUE #658: [📝 Docs]: Simplify examples on getting started page

Emerged out of the original issue #653.
Page - [getting-started-step-by-step](https://json-schema.org/learn/getting-started-step-by-step)

This issue tracks the progress on having simpler workings while explaining `required` keys to a new user onboarded.

Currently - explanation for required keys  of "productId", "productName" is kinda confusing. The docs does mentions that they are required but it can't be observed in the changes made.
It is only some time later that we see the line of code

```json
"required": [ "productId", "productName", "price]" 
```

Where it starts to make sense.

A better way to write it would be 

```js
  "properties": {
    "productId": {
      "description": "The unique identifier for a product",
      "type": "integer"
    }
  },

  "required": ["productId"]
```

i.e Update the required key as we go along adding new attributes. Not all at once (which is currently in the docs). As it makes a proper flow of what is written in text and simultaneously in code

Original comment - https://github.com/json-schema-org/website/issues/653#issuecomment-2056423433

### Code of Conduct

- [X] I agree to follow this project's Code of Conduct

chidiadi01:

In this commit, the following changes were made:

-  Added 'required' keyword to all code samples in the 'Define properties' section
-  Reworded and restructured the 'Define properties' section for better understanding, and removed the 'Add the properties object' sub-section because it was unnecessary.
-   Added an info box to give more information about optional properties

**Issue Number:**

-  Closes #658  

**Screenshots/videos:**
Screenshot of the reworded section

### DIFF
--- a/pages/learn/getting-started-step-by-step.md
+++ b/pages/learn/getting-started-step-by-step.md
@@ -90,83 +90,61 @@ In JSON Schema terminology, `$schema` and `$id` are [schema keywords](https://js
 
 ## Define properties[#define]
 
-This section adds the `properties` keyword. In JSON Schema terms, `properties` is a [validation keyword](https://json-schema.org/draft/2020-12/json-schema-core.html#section-10.3.2.1). When you define `properties`, you create an object where each property represents a key in the JSON data that’s being validated. You can also specify which properties defined in the object are required.
+`Properties` is a [validation keyword](https://json-schema.org/draft/2020-12/json-schema-core.html#section-10.3.2.1) used to define the keys expected in the json document. Here, we expect a JSON document used to define products. The keys expected are the `productId`, `productName`, `price`, and `tags`. These keys would have corresponding values of which the `productId` should have a value of integer type, the `productName` should have a string type value, the `price` key should have a number type value, and the `tags` should have an array as value. These all make up the `Product` object.
 
-### Add the properties object[#properties]
-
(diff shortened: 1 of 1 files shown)


chidiadi01:
> There needs to be discussion as to what action is needed before a PR is submitted.  There is no discussion on that issue.
> 
> 

That means I have to wait for @bhavukkalra to respond?

bhavukkalra:
Hey No Actually. You don't need to wait for me .We did had a discussion on this change was agreed upon that we should be adding diff changes to the Json entries. I am already in the process of completing that change. Works on my local by getting inconsistent results. 

Also, I think we use Info box currently to differentiate amongst different draft versions. Like If a certain keyword is not/is available in the draft and adding that as a generic info here could create confusion amongst readers.

I think that's what greg was trying to instantiate as well that a discussion should happen on what changes are currently in sync and which are not. So that a PR could be accurately reviewed

gregsdennis:
@chidiadi01 what I mean is that there's no discussion in the linked issue.  It's just a proposal.  There shouldn't be a PR from just a proposal. The issue needs to be discussed first to verify that s change needs to actually happen. 

The discussion that @bhavukkalra is taking about happened only this week during the OCWM call, which was after I left that comment.

chidiadi01:
Oh...ok. Noted.

benjagm:
Closing this PR as per https://github.com/json-schema-org/website/pull/671#issuecomment-2080053882

PR #675, opened 2024-04-22
Add section on schema resource and compound schema resource

Temidayo32:

**Issue Number:**

-  Related to #266 

**Summary**

As discussed in #266, this PR adds two sections to the glossary page.

No

### DIFF
--- a/pages/learn/glossary.md
+++ b/pages/learn/glossary.md
@@ -14,6 +14,12 @@ If you encounter a term you wish were defined here, please feel free to [file an
 
 The entries on this page can be linked to via anchor links (e.g. `https://json-schema.org/learn/glossary.html#vocabulary`) when sharing a definition with others.
 
+### compound schema resource
+
+A compound schema resource is a JSON document which has multiple embedded [JSON schema resources](#schema-resource). It is important to note that for a schema to be embeddable it must define a `$id` keyword, which is used to hold its unique identifier. These embedded schema resources collectively define various aspects or features of the overall schema.
+
+Compound schema resource is vital when working with multiple schema resources. By [bundling](../understanding-json-schema/structuring#bundlingbundling) these schema resources into a single schema document, distribution is made easier.
+
 ### dialect
 
 A cohesive collection of [keywords](#keyword) available for use within a schema, often representing a use-case specific single release of the JSON Schema specification.
@@ -106,6 +112,13 @@ Strictly speaking, according to the specification, schemas are themselves [JSON 
 
 In recent [drafts](#draft) of the specification, a schema is either a JSON object or a JSON boolean value.
 
+#### schema resource
+
+A schema resource is a [schema](#schema) which is canonically identified by an absolute URI. This implies all root/parent schemas are schema resources as they can be identified by an absolute URI. 
+
+The URI serves as a canonical identifier for the schema. This allows it to be referenced and reused in other schemas. Consequently, all schema resource must maintain semantic clarity, without which reusability will be difficult. 
+
+
 ### subschema
 
 A [schema](#schema) which is itself contained within a surrounding parent schema.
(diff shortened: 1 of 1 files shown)


pages/learn/glossary.md
```diff
@@ -106,6 +112,13 @@ Strictly speaking, according to the specification, schemas are themselves [JSON
 
 In recent [drafts](#draft) of the specification, a schema is either a JSON object or a JSON boolean value.
 
+#### schema resource
+
+A schema resource is a [schema](#schema) which is canonically identified by an absolute URI. This implies all root/parent schemas are schema resources as they can be identified by an absolute URI.
```
gregsdennis, line 117: > A schema resource is a [schema](#schema) which is canonically identified by an absolute URI.

This is just taken verbatim from [Core 4.3.5](https://json-schema.org/draft/2020-12/json-schema-core#section-4.3.5).  The idea of this glossary is to make things a bit more plain.

> This implies all root/parent schemas are schema resources as they can be identified by an absolute URI.

This isn't precisely true.

```json
{ "type": "object" }
```

is a root schema, but it's not a resource because it's not identifiable.

---

We should say that a schema resource is any schema or subschema which carries an `$id` keyword and then re-iterate the specification's suggestion that all root schemas should carry an `$id` keyword.

pages/learn/glossary.md
```diff
@@ -106,6 +112,13 @@ Strictly speaking, according to the specification, schemas are themselves [JSON
 
 In recent [drafts](#draft) of the specification, a schema is either a JSON object or a JSON boolean value.
 
+#### schema resource
+
+A schema resource is a [schema](#schema) which is canonically identified by an absolute URI. This implies all root/parent schemas are schema resources as they can be identified by an absolute URI. 
+
+The URI serves as a canonical identifier for the schema. This allows it to be referenced and reused in other schemas. Consequently, all schema resource must maintain semantic clarity, without which reusability will be difficult.
```
gregsdennis, line 119:
```suggestion
The value in the `$id` keyword should resolve to an absolute URI, which serves as a canonical identifier for the schema. This allows it to be referenced and reused in other schemas. 
```

> Consequently, all schema resource must maintain semantic clarity, without which reusability will be difficult.

I'm not sure what you mean by "semantic clarity".

pages/learn/glossary.md
```diff
@@ -14,6 +14,12 @@ If you encounter a term you wish were defined here, please feel free to [file an
 
 The entries on this page can be linked to via anchor links (e.g. `https://json-schema.org/learn/glossary.html#vocabulary`) when sharing a definition with others.
 
+### compound schema resource
+
+A compound schema resource is a JSON document which has multiple embedded [JSON schema resources](#schema-resource). It is important to note that for a schema to be embeddable it must define a `$id` keyword, which is used to hold its unique identifier. These embedded schema resources collectively define various aspects or features of the overall schema.
+
+Compound schema resource is vital when working with multiple schema resources. By [bundling](../understanding-json-schema/structuring#bundlingbundling) these schema resources into a single schema document, distribution is made easier.
```
gregsdennis, line 21: > Compound schema resource is vital when working with multiple schema resources.

I'm not sure this is true.  A system can happily work with every resource in its own document.  Ideally, this is the way it should be done, but that's not always practical.

Also bundling, as described by that link, isn't the only reason to embed resources.  In an OpenAPI document block, each schema could have its own identifier, making it a distinct resource which could be `$ref`'d from anywhere.

pages/learn/glossary.md
```diff
@@ -117,7 +130,7 @@ For example, the `not` keyword takes a subschema value and inverts its result, s
 Some subschemas may appear in more complex nested locations within a parent schema.
 The `allOf` keyword, for instance, takes an array of multiple subschemas and succeeds whenever all of the subschemas do individually.
 
-Whether something that otherwise *appears* to be a schema (based on its contents) actually *is* a subschema can be misleading at first glance without context or knowlege about its location within the parent schema.
+Whether something that otherwise *appears* to be a schema (based on its contents) actually *is* a subschema can be misleading at first glance without context or knowledge about its location within the parent schema.
```
gregsdennis, line 133: The spelling fix is appreciated, but the wording here is weird.  Can we clean this up while we're here?

gregsdennis (changes requested):

pages/learn/glossary.md
jdesrosiers, line 117: > A schema resource is a [schema](#schema) which is canonically identified by an absolute URI.

I'd drop the word "canonically". It's just techno-babel that's not necessary to describe the concept to lay-people.

jdesrosiers: Greg, I think you've uncovered an inconsistency in the spec. I think that your example has to be considered a Schema Resource. Here's a slightly more complex schema to explain why.

```json
{
  "type": "object",
  "properties": {
    "foo": { "$ref": "https://example.com/schema/foo" }
  },
  "$defs": {
    "foo": {
      "$id": "https://example.com/schema/foo",
      "type": "string"
    }
  }
}
```

The root schema is one of the Schema Resources that makes up this Compound Schema Document, but it doesn't have an identifier. If we don't consider the root schema a Schema Resource, then this wouldn't be a Compound Schema Document, which doesn't make sense. There has to be an exception so that the root schema doesn't need to have `$id`.

pages/learn/glossary.md
jdesrosiers: In addition to my :+1:  on Greg's suggested change, I suggest dropping the word "canonical". The lay-person doesn't know that word and it's not necessary to convey the concept.

jdesrosiers (changes requested):

benjagm:
We are closing this PR due to inactivity. Thanks everyone who was part of the discussion.

PR #677, opened 2024-04-25
Add validation result and JSON pointer in the glossary

starlightknown:
As discussed in the [issue](https://github.com/json-schema-org/website/issues/266). This PR adds two very similar sections to the [glossary](https://json-schema.org/learn/glossary) section.

**Issue Number:**

-  Related to #266  
- 
**Screenshots/videos:**

**Summary**

This PR is a part of the individual PR's getting created for completing the `glossary `page.

no

### DIFF
--- a/pages/learn/glossary.md
+++ b/pages/learn/glossary.md
@@ -55,6 +55,10 @@ The essence of JSON Hyper-Schema lies in its ability to define links and actions
 
 In other words, JSON Hyper-Schema extends JSON Schema by introducing features for creating hypermedia controls. This facilitates the creation of interactive APIs and ensures compatibility with existing JSON HTTP APIs, maintaining a seamless integration. It adds a layer of interactivity to JSON documents, making it easier to interact with remote JSON resources.
 
+### JSON pointer
+
+JSON Pointer is a format for data exchange between systems, defining a string for accessing and modifying values within a JSON document. By using JSON Pointer, we can precisely reference and retrieve specific values within a JSON document. The [subschema](#subschema) is typically identified via a JSON Pointer, or equivalent syntax, allowing for targeted validation and manipulation of data within the JSON document. 
+
 ### keyword
 
 A property appearing within a [schema](#schema) object.
@@ -136,6 +140,10 @@ The following are considered tools:
 - a UI form generator
 
 Applications which use JSON Schema internally without exposing that functionality in some way, for example, validating configuration files or web requests, are not considered tooling.
+
+### validation result
+
+The validation result in the context of JSON Schema refers to the outcome of applying the entire JSON Schema to the entire instance document. This application should yield a boolean assertion, indicating whether the instance document conforms to the rules and constraints specified in the schema. The validation result signifies whether the instance document passes or fails validation against the [schema](#schema).
 
 ### vocabulary
 
--- a/pages/learn/glossary.md
+++ b/pages/learn/glossary.md
@@ -57,7 +57,7 @@ In other words, JSON Hyper-Schema extends JSON Schema by introducing features fo
 
 ### JSON pointer
 
-[JSON Pointer](https://datatracker.ietf.org/doc/html/rfc6901) is a format for data exchange between systems, defining a string for identifying locations within a JSON document. It provides a means to precisely reference specific parts of the JSON document for retrieval or manipulation. The [subschema](#subschema) is typically identified via a JSON Pointer, or equivalent syntax, allowing for targeted validation of data within the JSON document.
+[JSON Pointer](https://datatracker.ietf.org/doc/html/rfc6901) is a string syntax for identifying a value at a specified location within a JSON document. It serves to precisely reference specific parts of the document for retrieval or manipulation. A subschema is often identified using a JSON Pointer, specifying its location within the containing resource.
 
 ### keyword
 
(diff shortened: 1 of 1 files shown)


pages/learn/glossary.md
```diff
@@ -55,6 +55,10 @@ The essence of JSON Hyper-Schema lies in its ability to define links and actions
 
 In other words, JSON Hyper-Schema extends JSON Schema by introducing features for creating hypermedia controls. This facilitates the creation of interactive APIs and ensures compatibility with existing JSON HTTP APIs, maintaining a seamless integration. It adds a layer of interactivity to JSON documents, making it easier to interact with remote JSON resources.
 
+### JSON pointer
+
+JSON Pointer is a format for data exchange between systems, defining a string for accessing and modifying values within a JSON document. By using JSON Pointer, we can precisely reference and retrieve specific values within a JSON document. The [subschema](#subschema) is typically identified via a JSON Pointer, or equivalent syntax, allowing for targeted validation and manipulation of data within the JSON document.
```
gregsdennis, line 60: It'd be good to link to the JSON Pointer spec (RFC 6901).

Also, JSON Pointer merely identifies a location within a document. It doesn't (itself) access or modify values, though it _is_ used by other standards to accomplish those tasks.

gregsdennis (changes requested):

pages/learn/glossary.md
gregsdennis: Thanks for adding the link, but the rest has not been addressed.  JSON Pointer does not modify documents.  It only serves to indicate a location within a document.

gregsdennis (approved):
Much better. Thank you. 

Would like another pair of eyes on this from the team.

pages/learn/glossary.md
```diff
@@ -55,6 +55,10 @@ The essence of JSON Hyper-Schema lies in its ability to define links and actions
 
 In other words, JSON Hyper-Schema extends JSON Schema by introducing features for creating hypermedia controls. This facilitates the creation of interactive APIs and ensures compatibility with existing JSON HTTP APIs, maintaining a seamless integration. It adds a layer of interactivity to JSON documents, making it easier to interact with remote JSON resources.
 
+### JSON pointer
+
+[JSON Pointer](https://datatracker.ietf.org/doc/html/rfc6901) is a format for data exchange between systems, defining a string for identifying locations within a JSON document. It provides a means to precisely reference specific parts of the JSON document for retrieval or manipulation. The [subschema](#subschema) is typically identified via a JSON Pointer, or equivalent syntax, allowing for targeted validation of data within the JSON document.
```
notEthan, line 60: > JSON Pointer is a format for data exchange between systems

I don't think "format for data exchange between systems" is a good description of JSON pointer. It just points to a location in a document - I don't see any data exchange between any systems (it might be used as part of such a thing, but that's not what JSON pointer is).

> JSON Pointer, or equivalent syntax

I'm not sure what "equivalent syntax" would be. A subschema could be identified by an anchor, but that's not equivalent syntax.

---

The JSON Pointer RFC has a pretty good start at a definition:

> JSON Pointer defines a string syntax for identifying a specific value within a JavaScript Object Notation (JSON) document.

I'd say more about referencing a _location_ in the document (as this PR does) - combining this PR and the RFC I'd end up with something like

> [JSON Pointer](https://datatracker.ietf.org/doc/html/rfc6901) is a string syntax for identifying a value at a specified location within a JSON document. A [subschema](#subschema) is often identified using a JSON Pointer specifying its location in the containing resource.

pages/learn/glossary.md
```diff
@@ -137,6 +141,14 @@ The following are considered tools:
 
 Applications which use JSON Schema internally without exposing that functionality in some way, for example, validating configuration files or web requests, are not considered tooling.
 
+### validation result
+
+The validation result in the context of JSON Schema refers to the outcome of applying the entire JSON Schema to the entire instance document. This application should yield a boolean assertion, indicating whether the instance document conforms to the rules and constraints specified in the schema. The validation result signifies whether the instance document passes or fails validation against the [schema](#schema).
```
notEthan, line 146: A validation result can be much more than a boolean assertion - there's a whole set of output formats that can be the result of validation, of which a boolean is the simplest.

pages/learn/glossary.md
```diff
@@ -137,6 +141,14 @@ The following are considered tools:
...
+
+#### See also
+
+[Get started with JSON schema](../blog/posts/get-started-with-json-schema-in-node-js.md)
```
notEthan, line 150: I don't think this 'see also' belongs on the 'validation result' entry - it might be appropriate to mention in the intro, but probably most, if not all, of this glossary relates to getting started.

pages/learn/glossary.md
gregsdennis: I think this is fine for "validation" result.  I'd consider the rest of the output to be "evaluation" results.  Not sure if we need to split that particular hair here, though.

benjagm (approved):
Looks great to me!

pages/learn/glossary.md
jdesrosiers: When I use the term "validation result", I always mean just the boolean result. I wouldn't be against using an expanded definition here, but I've never used it that way in the past.

gregsdennis (approved):

gregsdennis:
I'm going to take this as is.  If we want to improve upon it later, we can open a new PR.

Issue #549, opened 2024-03-16
✨ Enhancement: Rebuild the getting started guide


benjagm:
### Is your feature request related to a problem? Please describe

We are aware that learning JSON Schema is not easy and we also know that our docs needs a lot of improvement. But now, with the new feedback component we are even more clear about the areas that needs improvement and the getting started guide is clearly one of them.

After long consideration, I think the best approach is proceeding with a first iteration consisting on changing the section [Validating JSON data against the schema](https://json-schema.org/learn/getting-started-step-by-step#validate).

We don't maintain or promote any validator over others and in addition we are not integrating any validator. This is great to keep our neutrality but it adds friction into new users journey. We are going to keep that neutrality but it will be great to allow the users feel the HAHA moment of getting a json message validated against a schema.

My proposal is to create a component that provides different valid and invalid examples for different uses cases following this prototype:

<img width="380" alt="Screenshot 2024-04-18 at 15 51 05" src="https://github.com/json-schema-org/website/assets/40007659/a8fa4a73-5900-4fe0-9c76-d996da1bf8a7">

The default example shown by the component is the one used for getting started. Considering that the examples are pre-recorded and we know the validation result. 

1. Initially the component shows the default schema instance.  (this will be paremetrized in a yaml file)
2. The user can select 2 or 3 different instances, valid and invalid and we show the result and the explanation why showing a green check when is valid and red cross when is invalid.
3. We'll provide some other schema examples with instances so the users can see it working in different scenarios. We add simple and other more complex examples.
4. We provide a download button to allow to download the schema and the selected instance so the user can try it with the validator of their choice.

The idea is let the users see the power of the json schema validation and show other real life examples to inspire them.

benjagm:
Some examples: 
 * https://docs.astro.build/en/getting-started/
 * https://docs.github.com/en/get-started
 * https://gohugo.io/getting-started/quick-start/

VivekJaiswal18:
Hello @benjagm, can I help you with this in any way.

benjagm:
Thanks Vivek. I still need to find some time to create a prototype with what I have in mind.

crampussm:
Hey @benjagm can I work on this?

benjagm:
> Hey @benjagm can I work on this?

My apologies for the late response. I think @DhairyaMajmudar is working on this. Isn't it?

DhairyaMajmudar:
Yes I am currently working on this.

PR #687, opened 2024-05-10
Feat: Rebuilding getting started component

DhairyaMajmudar:

Rebuilding getting started guide

**Issue Number:**

-  Related #549 

**Screenshots/videos:**
N/A

N/A

**Summary**
Getting stated guide added. 

No


DhairyaMajmudar:
I want the data to be filled within data files.

benjagm (changes requested):
This is very very close! Congrats!

Left some suggestions to change texts and also found that the default instance is not loading properly.

components/GettingStarted.tsx
```diff
@@ -0,0 +1,208 @@
...
+      valid: false,
+      file: '/data/getting-started-examples/instances/default-ko.json',
+      details: 'This is an invalid JSON instance for the provided JSON Schema',
+    },
+  ]);
+
+  const [details, setDetails] = useState([
+    'This is a valid JSON instance for the provided JSON Schema',
+    true,
+  ]);
```
benjagm, line 49: Hi Dhairya, all those methods should disappear from here as they are overriding the information available in the json files. The idea is this component to read getting-started-examples.json and find all this information.

Probably this is here because you was testing something, but this should disappear and the behaviour should be totally data driven with getting-started-examples.json as source.

benjagm (changes requested):
Left a comment regarding the GettingStarted component. There is some hardcoded logic that is not supposed to be there.

components/GettingStarted.tsx
DhairyaMajmudar: Hi Benja I have kept this data here to initialize the state variables, if they are removed it will further affect other state variables and its functioning

benjagm (approved):
Looks great to me. I added the last changes and now I am asking for some more reviews.

PR #704, opened 2024-05-21
Update "Validating OpenAPI and JSON Schema" post

jdesrosiers:
I saw a [question](https://stackoverflow.com/questions/78424948/openapi-validator-that-supports-a-custom-dialect) on StackOverflow that asked about validating OpenAPI documents with a custom JSON Schema dialect. I pointed them toward this blog post that answers part of their question, but I realized there are some brief code examples that use an old version of `@hyperjump/json-schema`, so I wanted to update those examples.

### DIFF
--- a/pages/blog/posts/validating-openapi-and-json-schema.md
+++ b/pages/blog/posts/validating-openapi-and-json-schema.md
@@ -65,42 +65,33 @@ any dialect you choose to use in your OpenAPI 3.1 documents.
 
 ### Examples
 
-These examples use the [@hyperjump/oas-schema-validator](https://www.npmjs.com/package/@hyperjump/oas-schema-validator)
-to validate OpenAPI documents. It's just an extension of [@hyperjump/json-schema](https://www.npmjs.com/package/@hyperjump/json-schema)
-with support for the OpenAPI 3.1 dialect and all the OpenAPI 3.1 schemas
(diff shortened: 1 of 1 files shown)


gregsdennis (approved):

benjagm (approved):
Nice job with this article! I learned a lot while reading it. I think there is range for more posts in this intersection between OpenAPI and JSON Schema.

Looks great to me. My only thought regarding the format of the article: The article ends after the last sub-section of the "Supporting multiple dialects" with a cold end for what was a great article. My suggestion is adding a closing paragraph or closing thoughts section to avoid that cold end.

gregsdennis:
Do we want to change the spirit of the article? I think this change is just updating for relevant code samples.

benjagm:
> Do we want to change the spirit of the article?

It is all right from my side. I approved the PR but also shared my opinion just in case it resonates to Jason.

jdesrosiers:
@benjagm is absolutely right that this post would be better if it had a proper conclusion. I remember writing it and couldn't think of a way to end it. I don't have a problem with updating an old post, but I'm a bit overloaded at the moment and don't want to delay releasing this update until I have time to come back to this and work on adding a conclusion. I'll put that on my list and hopefully get to it sometime in the near future.

PR #715, opened 2024-05-29
Update online JSON Schema Validator to support 2020-12

JamesNK:
Update the doc to say that https://www.jsonschemavalidator.net/ supports 2020-12.

Is this all that is required to update https://json-schema.org/implementations?

### DIFF
--- a/data/validator-libraries-modern.yml
+++ b/data/validator-libraries-modern.yml
@@ -476,12 +476,12 @@
       last-updated: "2022-08-31"
     - name: JSON Schema Validator
       url: https://www.jsonschemavalidator.net/
-      date-draft: [2019-09]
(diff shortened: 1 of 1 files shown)


JamesNK:
Json.NET Schema (library it is built on top of) also supports 2020-12 but only in a pre-release version. Should it be updated now or wait until a final version is public?

gregsdennis:
@JamesNK we'd prefer to update here once you're out of pre-release.

Also, I'm having trouble hitting https://www.newtonsoft.com/, but it might be something at my end (I'm in Spain right now).

Thanks for the update.

JamesNK:
Ok. This PR is fine to merge now.

I’ll make another PR in the future for the standalone library.

gregsdennis (approved):

PR #723, opened 2024-05-30
fix: updated icon `alt`s

1p22geo:

Update the alt text of icons so that the page is more accessible

**Issue Number:**

I don't think there is any

**Screenshots/videos:**
This is how the page looks like if you disable JS in the browser.
image

I don't think it is relevant.

**Summary**

Somehow all icons had an `alt` text of `eye icon` even if they were different icons.

no.


benjagm (approved):
Thanks a lot for this contribution @1p22geo !!! Much appreciated!

PR #679, opened 2024-04-26
Restructure what is JSON Schema for clarity

Nickyshe:
**What kind of change does this PR introduce?**
This issue is for restructuring and rewriting what is Json schema page for clarity, consistent use of pronouns, revamp headings and ensure the tone is consistent throughout the documentation 

**Issue Number:**

-  Closes #670 

**Screenshots/videos:**

Screenshot (528)

**If relevant, did you update the documentation?**

Yes I updated the documentation

PR #655, opened 2024-04-15
Web release 3

benjagm:
This is the release 3 of the Documentation strategy containing:

 - New **welcome** page to the docs:
   * Issue: #421 
   * New page: https://web-release-3.website-2v2.pages.dev/overview/welcome
  
 - New **FAQ** page:
   * Issue: #190 
   * New page: https://web-release-3.website-2v2.pages.dev/overview/faq

 - New **Community** page:
    * Issue: #280
    * New page: https://web-release-3.website-2v2.pages.dev/community

- New **use cases** page:
   * Issue: #426 
   * New page: https://web-release-3.website-2v2.pages.dev/overview/case-studies

- New **case studies** page:
   * Issue: #440 
   * New page: https://web-release-3.website-2v2.pages.dev/overview/use-cases

PR #724, opened 2024-05-30
Adding feedback to improve the messaging and illustration of What is JSON Schema

benjagm:
**Summary**

After merging #679 we receive great feedback from @DavidBiesack and this PR is intended to address David's suggestions.

No


DavidBiesack (approved):

public/img/json_schema.png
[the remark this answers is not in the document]
benjagm: Thanks once again! I updated the illustration accordingly!

benjagm:
@Relequestual We have incorporated in this PR your feedback from #655 review.

DavidBiesack (approved):

pages/overview/what-is-jsonschema.md
```diff
@@ -2,50 +2,49 @@
...
+JSON validators consist of both open-source and commercial tools that implement the JSON Schema specification. Such tooling enables easy integration of JSON Schema into projects of any size.
 
 ![How JSON Schema works](/img/json_schema.svg)
 
 ## Benefits of JSON Schema for Developers
 
 JSON Schema empowers developers in the following ways:
 
-* **Structured Data Description**: JSON Schema allows developers to describe the structure, constraints, and data types of existing JSON formats.
+* **Structured Data Description**: JSON Schema empower developers to use JSON to describe the structure, constraints, and data types of existing JSON data.
```
Relequestual, line 23:
```suggestion
* **Structured Data Description**: JSON Schema empowers developers to use JSON to describe the structure, constraints, and data types of existing JSON data.
```

public/img/json_schema.svg
Relequestual, line 1: I think the intent in the original image was to show that the Schema is applied ot the instance. I agree that showing the validation result can be true or false is useful though.

pages/overview/what-is-jsonschema.md
```diff
@@ -2,50 +2,49 @@
 section: docs
 title: What is JSON Schema?
 ---
-JSON Schema is a declarative language for annotating and validating JSON documents' structure, constraints, and data types. It provides a way to standardize and define expectations for JSON data.
+JSON Schema is a declarative language for annotating and validating JSON instances' structure, constraints, and data types. It provides a way to standardize and define expectations for JSON data.
```
Relequestual, line 5: This is a key statement which it would be good to make sure the TSC sees and agrees on.

pages/overview/what-is-jsonschema.md
Relequestual, line 23: Not everyone uses JSON to define their JSON Schema. Some use YAML, others use functional libraries with currying to construct the schema. I don't think adding "use JSON to" makes sense here.

Relequestual (changes requested):
Mostly comments, however as noted I would like to give the TSC a chance to review some of these changes.

I'm not sure about the image change, but I'm not strongly against it, so it would be good to see how others perceive this.

pages/overview/what-is-jsonschema.md
Julian: I think "instances" is terminology precisely 0% of people understand who haven't read it somewhere in our docs (who seem likely to be 0% of people reading this page).

The consequence of that I think is:

* that word should be immediately interlinked to the glossary the first time it's used

but also

* I would personally not use it in the first sentence certainly. I'd use "JSON data".

pages/overview/what-is-jsonschema.md
```diff
@@ -2,50 +2,49 @@
...
 
-While JSON Schema provides the language, validating a JSON [instance](https://json-schema.org/learn/glossary#instance) against a [schema](https://json-schema.org/learn/glossary#schema) requires a JSON Schema [validator](https://json-schema.org/implementations#tools). The JSON validator checks if the JSON documents conform to the schema.
+While JSON Schema provides the language, validating a JSON [instance](https://json-schema.org/learn/glossary#instance) against a [schema](https://json-schema.org/learn/glossary#schema) requires a JSON Schema [validator](https://json-schema.org/implementations#validators). The JSON validator checks if the JSON documents conform to the schema.
 
-JSON validators are open-source and commercial tools that implement the JSON Schema specification. They make it very easy to integrate JSON Schema into projects of any size.
+JSON validators consist of both open-source and commercial tools that implement the JSON Schema specification. Such tooling enables easy integration of JSON Schema into projects of any size.
```
Julian, line 15: I don't personally know what the reason to stress that validators can be either open or closed source here. It's a thing true of any software. But the diff here makes it seem like you need *both* kinds simultaneously to me I think.

I'd delete the sentence entirely TBH, but alternatively "There are both open-source as well as closed-source tools which implement the specification" or something.

pages/overview/what-is-jsonschema.md
```diff
@@ -2,50 +2,49 @@
...
+* **Access to a Vibrant Tooling Ecosystem**: JSON Schema is supported by a diverse array of languages, libraries, and frameworks with community-driven [tools](https://json-schema.org/tools). This vibrant ecosystem enhances development productivity and provides resources for effective schema implementation and utilization.
 
 
 ## History of JSON Schema 
 
 
 JSON Schema dates back to the [first JSON Schema proposal](https://web.archive.org/web/20071026185150/http://json.com/json-schema-proposal/) submitted by Kris Zyp to [json.com](http://json.com) on October 2nd, 2007.
 
-The current version of JSON Schema is [2020-12](https://json-schema.org/draft/2020-12/release-notes), which represents the latest advancements and has expanded capabilities compared with the previous versions, `draft-04`, `draft-06`, and `draft-07`.
+The newest version of JSON Schema is [2020-12](https://json-schema.org/draft/2020-12/release-notes), which represents the latest advancements and has expanded capabilities compared with the previous versions, `draft-04`, `draft-06`, and `draft-07`.
```
Julian, line 47: Some way of ensuring this sentence stays up to date (via CI) seems like a nice issue to file. Alternatively having a `https://json-schema.org/latest/release-notes` link might be nice, though because we have plain markdown here that wouldn't fix the "problem" that the text will still say "2020-12"...

public/img/json_schema.svg
DavidBiesack: my original comment was that the (original) image is misleading - it implied a flow where an JSON schema is transformed and a new result  or somehow passes through the instance and then a value passes to the validator. I think in reality, both the instance and the schema are passed to the validator _separately_, and a validation result flows out.

pages/overview/what-is-jsonschema.md
DavidBiesack: "define expectations for JSON data" is also confusing - data does not have expectations. People (developers) have expectations.
Perhaps "It provides a standard way to define what JSON data is valid for applications".

gregsdennis: "define people's expectations for JSON data"

public/img/json_schema.svg
gregsdennis: Benja showed this to both @jdesrosiers and me while we were in Barcelona, and we both (independently) agreed that the new image is more correct.  A validator takes a schema and an instance and provides an output.  The specific architecture an implementation uses is beside the point.  In the end, both the schema and the instance are data; a schema doesn't _do_ anything.

pages/overview/what-is-jsonschema.md
DavidBiesack: I agree that terms should be defined on first use. A glossary link is less than ideal in an intro, but better than nothing. The point is, I think these pages should use terms _consistently_ and avoid defining aliases which can raise questions (especially if those aliases are not defined in the glossary).
The glossary defines "instance" but not "document" (it probably should, and "JSON data" as well,
and perhaps distinguish between them all.)  I think many people think a "document" means streamable content in a file (or web resource), as opposed to in-memory data (i.e. file.json vs. a parsed JSON value in a data structure in a programming language), which is why I suggest not using "document" here. "JSON data" is better but it does add an alias for "instance"...

gregsdennis: I also agree with Julian that "instances" might be a bit vernacular-y at this point.

```suggestion
> JSON Schema is a declarative language for defining structure, constraints, and data types for JSON data.
```

jdesrosiers: > structure, constraints, and data types

I feel like these words are all saying the same thing and using all of them is confusing. I'd just use "constraints" because JSON Schema isn't really about structure or data types. Those things are just a consequence of the constraints of the schema.

pages/overview/what-is-jsonschema.md
benjagm: Thanks for the clarification Ben!

pages/overview/what-is-jsonschema.md
benjagm: Thanks. I'll remove entirely the distinction of open - comercial validators.

pages/overview/what-is-jsonschema.md
benjagm: Great idea. I'll add that redirect and modify the link accordingly.

benjagm:
Thanks everyone!! I just pushed a new version addressing all the comments.

PR #752, opened 2024-06-08
Update Json.NET Schema to support 2020-12

JamesNK:
Follow up to https://github.com/json-schema-org/website/pull/715 now that it's publically released.


gregsdennis (approved):
You should look into adding your implementation to Bowtie.  (https://bowtie.report)

benjagm (approved):

PR #727, opened 2024-05-31
chore(package.json): Update dependencies and node version

1p22geo:
**What kind of change does this PR introduce?**

- Use node v20.9.0+ (or anything > 20.9.0) as specified in README instead of stricly requiring node == 20.0.0 (as it was in package.json)
- Update to use latest yarn
- Yarn is a package manager and not a dependency, it is not needed in runtime. Yarn installs the dependencies, so no need to mention it in `dependencies` in package.json

**Issue Number:**
There weren't any, as it is not *technically* a bug. Yet.

**Screenshots/videos:**

Sorry, none here.

**If relevant, did you update the documentation?**

Not relevant.

**Summary**

[why upgrade yarn](https://yarnpkg.com/migration/overview)

I know that `package.json` updates are stricly forbidden for this repository (there is even a CI workflow to fail PRs which changed it) but this is kinda important.

I know it will cause major pain, especially for CI runs, but I can help with migrating workflows to yarn 4.

**Does this PR introduce a breaking change?**

Yes.

PR #757, opened 2024-06-12
Continuation of PR 727 - chore(package.json): Update dependencies and node version

benjagm:
The PR #727 was merged into **web-node-20-9-0** branch to make some test of the node and yarn version in cloudflare.

Now that everything works this PR is intended to promote the change to the main branch.


jdesrosiers (commented):
`.yarn` and `.yarnrc.yml` shouldn't be checked in. It looks like you have `.yarn` in `.gitignore` but for some reason `.yarn/releases/yarn-4.2.2.2.cjs` is getting checked in.

I notice that there are changes to `styles/globals.css` in this PR. Is that a mistake? It doesn't seem to have anything to do with this PR.

jdesrosiers:
> It seems that .yarn/releases/. should be added to the repos as recommended in the yarn docs

Committing large binarys in git is considered bad practice. Committing external dependencies in your project is considered bad practice. So, I had to look into what was going on here. It looks like yarn has changed how they recommend installing yarn and you ended up using an out-dated approach.

The [current recommendation](https://yarnpkg.com/getting-started/install) is that `corepack` is used in conjunction with `"packageManager"` in package.json to automatically install and use the correct version of yarn. With this approach, `corepack` manages the binaries and there's nothing for us to check in or ignore (as it should be).

The older way of installing yarn installed the binary to the `.yarn/releases` folder and recommended that you check it in to git. The `.gitignore` is still configured to allow you to check in `.yarn/releases` for backward compatibility with the older approach. The new approach simply doesn't use that directory.

You seem to have ended up using this older install method instead of using `corepack`. I recommend removing everything you have related to yarn and doing a fresh install using corepack. Do `corepack enable` to ensure corepack is installed and enabled. Then all you have to do is run `yarn install` and everything will get installed automatically. You don't even need to do `yarn set version 4.2.2` because it will get the right version from the `"packageManager"` field in `package.json`. I've done this and can confirm that it works and doesn't include a `.yarn/releases` folder.

benjagm:
Closed in favour of #842

PR #782, opened 2024-06-24
Feat: Rebuilding getting started component (#687)

### LINKED ISSUE #549: ✨ Enhancement: Rebuild the getting started guide

benjagm:

Rebuilding getting started guide

**Issue Number:**

-  Closes #549 

**Screenshots/videos:**
N/A

N/A

**Summary**
Getting stated guide added. 

No


benjagm:
The new validation components of the getting started guide con tested here:

https://web-new-getting-started.website-2v2.pages.dev/learn/getting-started-step-by-step#validate-json-data-against-the-schema

DhairyaMajmudar:
I have added the download button which on click downloads the schema FIle data

benjagm:
Hi Dhairya. Can you please fix the conflict and make sure that the download button downloads schema and instance in a zip file with the original names?

benjagm (commented):
Great job @DhairyaMajmudar. Let's merge this!

PR #792, opened 2024-07-08
Feat: Adding Side ToC

DhairyaMajmudar:

Adds side ToC

**Screenshots/videos:**

Screenshot from 2024-07-08 12-02-29

N/A

**Summary**
Side ToC added

### DIFF
--- a/pages/learn/[slug].page.tsx
+++ b/pages/learn/[slug].page.tsx
@@ -4,9 +4,10 @@ import StyledMarkdown from '~/components/StyledMarkdown';
 import { getLayout } from '~/components/Sidebar';
 import getStaticMarkdownPaths from '~/lib/getStaticMarkdownPaths';
 import getStaticMarkdownProps from '~/lib/getStaticMarkdownProps';
-import { Headline1 } from '~/components/Headlines';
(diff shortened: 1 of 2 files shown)


pages/learn/[slug].page.tsx
```diff
@@ -4,9 +4,10 @@ import StyledMarkdown from '~/components/StyledMarkdown';
 import { getLayout } from '~/components/Sidebar';
```
Comment (benjagm) [inline pages/learn/[slug].page.tsx:4]: Hi Dhairya. Great work adding the TOC on this page. However, we must add it to all the docs pages using the sidebar. This means, reference, spec, and the rest of the pages.

benjagm (changes requested):
Hi Dhairya. Great work adding the TOC on this page. However, we must add it to all the docs pages using the sidebar. This means, reference, spec, and the rest of the pages.

pages/learn/[slug].page.tsx
Comment (benjagm) [reply pages/learn/[slug].page.tsx:4]: @DhairyaMajmudar have you fixed this?

benjagm:
Hi Dhairya, I just thought that this side TOC should not require to be added page by page, instead it should be part of the standard layout, probably as part current sidebar component.

PR #838, opened 2024-08-01
Add workflow to update landscape tooling

DarhkVoyd:
This PR aims to add a workflow, to track changes to the tooling data file and update tools in our landscape tooling category when tooling data changes.

NOTE: We need to add a fine-grained personal access token named `PERSONAL_ACCESS_TOKEN`, with `write` access to `Actions` of `landscape` repository to have appropriate authentication.

**Issue Number:**

-  Closes #___ 
-  Related to #___ 
-  Others? 

**Screenshots/videos:**

**Summary**

### DIFF
--- a/.github/workflows/trigger-update-landscape-tooling.yml
+++ b/.github/workflows/trigger-update-landscape-tooling.yml
@@ -1,0 +1,26 @@
+name: Trigger Update Landscape Tooling Workflow
+
+on:
+  push:
+    paths:
+      - data/tooling-data.yaml
+    branches:
+      - main
+  workflow_dispatch:
+
+
+jobs:
+  trigger-workflow:
+    runs-on: ubuntu-latest
+
+    steps:
+      - name: Trigger update landscape repository workflow
+        env:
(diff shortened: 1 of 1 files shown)


aialok:
Why not to use our json-schema-bot token instead of PAT?

benjagm:
You are totally right! My bad! Can you update it DV?

DarhkVoyd:
Hey, sure! I didn't have access to existing token names so I just assumed a new token. Can you provide the name of the existing token? Moreover, please ensure that the token has `read and write` actions permissions.

aialok:
Below is only step need to add to create token and using id you can use that token in other step.





You can also checkout linkchecker workflow, where are using that bot to create  issue.

We might not have permission to trigger workflow but we can update permissiom for it easily.

DarhkVoyd:
Got it, Thanks! I think there was some miscommunication from my side, its not a personal token per say, it was just any token with actions permission I just named it "Personal Access Token"

aialok:
> Got it, Thanks! I think there was some miscommunication from my side, its not a personal token per say, it was just any token with actions permission I just named it "Personal Access Token"
> 

Good part about github app token is that it get destroy after the action finish : )

PR #843, opened 2024-08-03
fix(security): PR Previews from Forks

AdrianGonz97:
## The Issue

Hey! I'm the maintainer of the Github Action [refined-cf-pages-action](https://github.com/AdrianGonz97/refined-cf-pages-action) and we were recently made aware of a security vulnerability regarding the `pull_request_target` workflow event where it's possible to leak secrets (including the CF credentials used in the action) through a Github Actions exploit when running untrusted code. Unfortunately, our previous recommendation for the setup of the _PR Previews from Forks_ feature included the use of this workflow event type without PR approvals, which [has now been updated](https://github.com/AdrianGonz97/refined-cf-pages-action#enabling-pr-previews-from-forks). 

I'm going around to all of the dependents of the action that are using the _PR Previews from Forks_ feature to apply the fix.

## The Fix

We've come up with an alternate method that is safer to use and will resolve this issue entirely. 

Rather than using `pull_request_target`, which runs in a privileged environment (meaning that repository secrets can be used in it), previews will now be deployed in two stages:

1. The first stage (`build-preview.yml`) will use the `pull_request` event, which runs in an unprivileged environment, making it safe to run untrusted code. The site will be built in this stage and the build directory will be uploaded to Github as an [artifact](https://docs.github.com/en/actions/writing-workflows/choosing-what-your-workflow-does/storing-workflow-data-as-artifacts).
2. The second stage (`deploy-preview.yml`) will use the `workflow_run` event, which runs in a privileged environment where its only job will be to download the build artifact and then run the `refined-cf-pages-action` action to upload the build files to Cloudflare Pages for preview deployment.

## Next Steps

1. Since secrets can't be used in `pull_request` workflow events, `NEXT_PUBLIC_ALGOLIA_API_KEY` and `NEXT_PUBLIC_ALGOLIA_APP_ID` will need to be moved to repository variables instead.

---

And that's it, thanks for your time!


gregsdennis:
Thanks for putting this up, @AdrianGonz97!  It's always appreciated with a maintainer looks after their users.

aialok (approved):
LGTM :rocket: 

Thank you @AdrianGonz97  for this : )

benjagm (approved):
Thanks a lot @AdrianGonz97 !! We really appreciate that you took the time to help us with the required changes.

Issue #780, opened 2024-06-24
✨ Enhancement: Integrate Bowtie Reports


DarhkVoyd:
**Description:**

Integrate Bowtie to provide users with detailed benchmarks and comparisons of JSON Schema implementations, enhancing the data-driven nature of the tooling page.

**Major Tasks:**
- [x] **Integration**
  - Connect to the Bowtie to fetch benchmark data.
  
- [x] **Data Display**
  - Design and implement UI components to display Bowtie benchmarks within the tooling table such as passing badges.
  
- [x] **Data Updates**
  - Implement a mechanism to ensure that benchmark data is regularly updated and reflects the latest reports.
  
- [x] **User Interaction**
  - Provide interactive elements (e.g., tooltips, modals) to display detailed benchmark information on user interaction.

Issue #716, opened 2024-05-29
✨ Enhancement: Build Data-Driven Tooling Page


DarhkVoyd:
## Description:

This issue will serve as the parent issue for tracking the progress of the GSoC 2024 project focused on building a data-driven tooling page for the JSON Schema. 

## Epic Releases and Tasks:

- [x] #717
   
- [x] #722

- [x] #779

- [x] #780

- [x] #777

- [x] #776

- [x] #778

- [ ] **Dashboards**

Resolves: #160

### Are you working on this?

Yes

benjagm:
Looks great!

Issue #850, opened 2024-08-05
🐛 Bug: Some SVG images are not loading correctly in dark mode on the case studies page.


aialok:
### Describe the bug

- Some SVG images are not loading in correct theme in dark mode on the case studies page.
-  https://json-schema.org/overview/case-studies

### Steps To Reproduce

- Visit https://json-schema.org/overview/case-studies

### Expected Behavior

- It should render correctly in both light and dark mode.

### Screenshots

[Screencast from 2024-08-05 18-22-17.webm](https://github.com/user-attachments/assets/bee636ff-5bd1-4bcb-ad6b-c520de119ca4)

### Device Information [optional]

```markdown
- OS:
- Browser:
- version:
```

### Are you working on this issue?

No

### Do you think this work might require an [Architectural Decision Record (ADR)]? (significant or noteworthy)

No

bharathz8:
Hi @aialok 

I am interested in contributing to this issue. can I start to work on it ?

Thank you!

aialok:
> They are loading correctly. The SVGs we are using are black. That is the reason they are not visible after switching to dark mode.

Yes, they are loading correctly, but we need to use dark theme SVGs for some images.

harshsbhat:
Sure, I can work on this. I don't need an assignment. But want to know that is it discussed and ready to work.

Michael-Obele:
Hi @benjagm is this issue still available?

harshsbhat:
@Michael-Obele At least I am not working on this issue.

benjagm:
We can use the white version of the log when in dark mode. We have this already implemented on the landing page.

benjagm:
Can we assign this issue to you Michael?

Michael-Obele:
> Can we assign this issue to you Michael?

Yes. I'll be glad to help.

DhairyaMajmudar:
Hi @Michael-Obele any update on this issue : )

Michael-Obele:
Yeah, I'll finish it up this week. Been a bit pressed with work.

Michael-Obele:
> Hi @Michael-Obele any update on this issue : )

So yeah, I'm done with the changes but I've been hitting errors when trying to commit (even to my own repo) so that's what's delaying me. I would love some help with this if you don't mind

Michael-Obele:
Nope, nevermind, I got it working, making a PR now.

Michael-Obele:
[...]

PR #858, opened 2024-08-12
[pages][tools]: Implement Tooling Analytics and Integrate Bowtie Report

### LINKED ISSUE #778: ✨ Enhancement: Implement Tooling Trends Analytics
**Description:**

Post user interactions and engagement patterns and analytics on the tooling page to identify trends and insights for further improvements.

**Major Tasks:**
- [x] **Analytics Integration**
  - Integrate Google Sheets Analytics for user interactions and traffic.
  - Set up custom events to monitor specific actions (e.g., filter usage, link clicks).

DarhkVoyd:
This PR aims to provide some bug fixes, clean code refactor, enhanced code and add new enhancements such as implementation of analytics for tooling trends for the tooling page and integrate bowtie report in the current design.

**Issue Number:**
-  Closes #778 #780 
-  Related to: #716  
-  Others? 

**Screenshots/videos:**

**Summary**


data/tooling-data.yaml
```diff
@@ -901,6 +901,8 @@
   supportedDialects:
     draft: [6, 7, 2019-09, 2020-12]
   toolingListingNotes: 'Powered by JsonSchema.Net in Blazor WASM for client-side validation'
+  landscape:
```
benjagm, line 904: Why we have a parent property called Landscape? I think the logo is an important property and we don't need to group them into "landscape". I suggest avoiding it and creating the logo as any other property.

next.config.js
```diff
@@ -5,7 +5,31 @@ const nextConfig = {
   pageExtensions: ['page.tsx'],
   images: {
     unoptimized: true,
-  }
+  },
+  webpack(config) {
```
benjagm, line 9: Why is this change required?

benjagm (changes requested):
Added some comments.

gregsdennis:
I would also like a chance to review this from a UX perspective.  I'm curious how the Bowtie integration looks. 

Still waiting for a good build so that a preview is generated.

data/tooling-data.yaml
DarhkVoyd: The data model is implemented from the json schema eco-system tooling self identification schema, since it already had a landscape property with a logo field, I utilised the exact same model as I understood that the logo is more relevant to the landscape project. If you would like to instead extend the self identification schema and add logo as a root property, let me know.

next.config.js
DarhkVoyd: This change is required to optimize how we manage and render icons across our website. Currently, we're fetching icons as images from the server at runtime, which leads to several inefficiencies:

1. **Unnecessary Network Requests**: Every icon is fetched as a separate image file, increasing the number of network requests. This impacts load times and server performance.
2. **Complex Theme Management**: To handle different themes, we're using multiple useEffect hooks to fetch and update icons based on the current theme. This approach is repetitive and adds unnecessary complexity to our code.

By switching to SVGR, we can embed SVGs directly as React components. This allows us to:

1. **Reduce Network Requests**: Icons are embedded at build time, eliminating the need to fetch them from the server during runtime.
2. **Simplify Theme Management**: We can easily manipulate icon styles natively with CSS, like using `dark:fill-black` to change the fill color, instead of relying on multiple useEffect hooks.

data/tooling-data.yaml
DarhkVoyd: For now, the landscape property only exists in the schema and type interface but I only added a single logo to test it out. So, we can still update our data model slights to move the logo field to the root properties instead.

next.config.js
DarhkVoyd: Such as #850, currently updating SVGs with theme would require `useTheme` and `useEffect` hooks, which again is repetitive and leads to inconsistencies sometimes. This would all be as simple as adding `dark:` classes to the svg components if we use svg's as react components and embed them during build time.

gregsdennis: Is this _necessary_ for this PR to work? If not, it should be in a separate change.

The effort is appreciated, but if it's unrelated it needs to be addressed as such.

DarhkVoyd: Kind of but I can change it! The theming with `useEffect` was a headache so I tried this as a fix. The icons in the PR depends on this but they are few so I can easily switch back. Lemme know!

gregsdennis: I'm okay if it's related.  I'll let @benjagm make the call.

DarhkVoyd: Sure! Just to be clear, this does not affect any existing svg bundling with webpack because all current svgs use public path of the deployed page to point to the resource and instead of javascript modules `import`.

data/tooling-data.yaml
benjagm: Ok

next.config.js
benjagm: Thanks everyone for the great discussions. Let's continue with the change here.

benjagm (approved):
I pushed some small changes to improve the raise issues/bowtie buttons.

Everything else looks great! Amazing job. It is great to see this working!

DarhkVoyd:
@benjagm Thanks! I have rebase the PR with latest changes, fixed conflicts and linting issues. The current redirect takes us back to the previous implementations page, should I go ahead and remove it?

benjagm:
> @benjagm Thanks! I have rebased the PR with latest changes, fixed conflicts and linting issues. The preview is not working as expected as the redirects take us back to the previous implementations page, should I go ahead and remove it?

Yes, We'll need to replace the old redirect available in the _redirects file:
Now we have:
`/tools /implementations 301`

We should replace it by:
`/implementations /tools 301`

DarhkVoyd:
Now, we just need to add the backend API to complete the integration.

benjagm:
> Still waiting for a good build so that a preview is generated.

Now we have the preview here: https://github.com/json-schema-org/website/pull/858#issuecomment-2294805144

benjagm:
> Now, we just need to add the backend API to complete the integration.

Let's do it!

DarhkVoyd:
I think we also need update the main nav link to `/tools`. Next router might not use the Cloudflare redirects.

benjagm:
> I think we also need update the main nav link to `/tools`. Next router might not use the Cloudflare redirects.

Yes. Better to change that link. You can also perform a search in the codebase looking for internal links pointing to "/implementations" because I am sure there will be some more.

gregsdennis:
What page is the change on?  The preview link just takes me to the home page.

DarhkVoyd:
@gregsdennis I was just updating redirects and internal links, you can now easily preview changes on the tooling page.

gregsdennis:
Is this Bowtie "integration" just a link to the Bowtie site?  From the description, I thought it'd be more.

DarhkVoyd:
@gregsdennis to utilise the space on the tooling table most optimally the tooling table has a outlink to bowtie for anyone interested, If you click on any tool it will display the tool with detailed information including bowie report and compliance.

DarhkVoyd:
I have realised that, currently we only fetch the latest bowtie report during the website build. So, if for some reason there is not a new build of the website for some time then we would still be displaying outdated report. Is this the expected behaviour or should I go ahead and add a workflow to instead keep an updated bowtie report in the data folder which runs on a regular interval such as every month?

benjagm (approved):
All good from my side. Thanks DV!!

benjagm:
> Is this Bowtie "integration" just a link to the Bowtie site? From the description, I thought it'd be more.

When you click over a tool, you see the details in a pop-up and when that tool is integrated with Bowtie you'll see all the badges.

benjagm (approved):
This is ready from my side. Great job!!

PR #863, opened 2024-08-15
Clarify that JSON Toolkit supports 2019-09 and 2020-12

jviotti:


benjagm (approved):
Thanks Juan and congratulations!!

PR #869, opened 2024-08-16
chore(github-actions): update contributor and project sync workflows

aialok:

- Since we have branch protection rules, automated commits can't be made. Therefore, both workflows will now create pull requests instead.

**Issue Number:**
- None

**Screenshots/videos:**

**Summary**

### DIFF
--- a/.github/workflows/sync-contributors.yml
+++ b/.github/workflows/sync-contributors.yml
@@ -46,14 +46,28 @@ jobs:
               fs.writeFileSync('community.json', JSON.stringify(data, null, 2));
             
         - name: Commit changes
-          env:
-            GITHUB_APP_TOKEN: ${{ steps.get_workflow_token.outputs.token }}
           run: |
             git config user.name "the-json-schema-bot[bot]"
             git config user.email "the-json-schema-bot[bot]@users.noreply.github.com"
             git add community.json
-            git diff --quiet && git diff --staged --quiet || (git commit -m "chore(community): update community.json" && git push "https://x-access-token:${GITHUB_APP_TOKEN}@github.com/${GITHUB_REPOSITORY}.git" HEAD:${GITHUB_REF#refs/heads/})
-
+            git diff --quiet && git diff --staged --quiet || (git commit -m "chore(community): update community.json")
+        
+          
+        - name: Create Pull Request
+          uses: peter-evans/create-pull-request@v6
+          with:
+            token: ${{ steps.get_workflow_token.outputs.token }}
+            title: 'chore(community): update community.json'
+            body: 'This PR updates the community.json file with the latest contributors data.'
+            branch: 'update-contributors'
+            base: 'main'
+            author: 'the-json-schema-bot <the-json-schema-bot[bot]@users.noreply.github.com>'
+            branch-suffix: timestamp
+            labels: 'sync-contributors'
+            assignees: 'the-json-schema-bot'
+            draft: false
+            signoff: true
+            add-paths: 'community.json'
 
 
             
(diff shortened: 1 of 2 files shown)


benjagm (changes requested):
Great job with this!

Could it be possible to add another step to approve and another one to merge?

aialok:
> Great job with this!
> 
> Could it be possible to add another step to approve and another one to merge?

I don't think it's possible to do this within the same workflow since we need to create the pull request first which take time, then approve it, and finally merge it. But lemme check.

However, it could be easily done with a separate workflow.

benjagm:
> However, it could be easily done with a separate workflow.

We'll need to do something to avoid the cumulation of unreviewed PRs.

aialok:
> > However, it could be easily done with a separate workflow.
> 
> We'll need to do something to avoid the cumulation of unreviewed PRs.

I will be creating another workflow to approved and merged the pr.

aialok:
Added auto approve and auto merge in the same workflow.

aialok:
For auto merge condition must  be true:
image


https://github.com/peter-evans/enable-pull-request-automerge?tab=readme-ov-file#conditions

gregsdennis:
> Added auto approve and auto merge in the same workflow.

This seems like a bad idea.  (Unless I'm misunderstanding what's being added.)

It seems like you're wanting an action to automatically approve and merge PRs?  Why?  PRs are intended for _review_... by humans.

aialok:
> > Added auto approve and auto merge in the same workflow.
> 
> This seems like a bad idea.  (Unless I'm misunderstanding what's being added.)
> 
> It seems like you're wanting an action to automatically approve and merge PRs?  Why?  PRs are intended for _review_... by humans.

Basically, both workflows work to fetch the latest contributor data and our roadmap project data and save it in a file. This allows us to use the data on our website, helping to reduce the number of API calls to GitHub by using the data directly from a JSON file.

And since we have branch protection rule we can't directly push to main branch. We need to create pull request first.

https://github.com/json-schema-org/website/pull/869#pullrequestreview-2244053294

gregsdennis:
> And since we have branch protection rule we can't directly push to main branch. We need to create pull request first.

Then maybe this repo isn't the best place to store the results of the data fetch.  Maybe we store it in another repo (or a gist) where only the bot has push rights (and maybe the project admins).  I use a [gist](https://gist.github.com/gregsdennis/28607f2d276032f4d9a7f2c807e44df7) to store some data for a custom badge on my repo ([action](https://github.com/json-everything/json-everything/blob/119788d83e3b016ba1fe8a0c5deb019b0193fad2/.github/workflows/dotnet-core.yml#L91-L100)).

Automation like this seems subversive to me.  When I encounter hacks like this, it usually means I've taken a wrong turn somewhere; that I'm doing something wrong.

benjagm:
> Automation like this seems subversive to me. When I encounter hacks like this, it usually means I've taken a wrong turn somewhere; that I'm doing something wrong.

Hi everyone here. This is the solution I suggested and I take responsibility. The fetched files will be used to build a future roadmap page and improve the current community page. We are storing the data to avoid doing synchronous API calls and avoid rate limit issues. I understand your concerns and I am happy to discuss alternatives. Gist seems like over-complicating things but I have never implemented it before. I know that AsyncAPI is doing something similar to us in their repository.

gregsdennis:
I expect you're looking at something like https://github.com/asyncapi/website/pull/3142?

If that's the direction you want to go, okay.  I've given my opinion.

benjagm (approved):
LGTM

PR #900, opened 2024-08-31
docs(implementations): added Schema generator dadav/helm-schema

outofrange:
This merge request adds [dadav/helm-schema](https://github.com/dadav/helm-schema) as a _Schema generator - From data_.

After finding [helm-docs](https://github.com/norwoodj/helm-docs) for creating human readable tables of Helm `values.yaml` files in Markdown or HTML tables, I was looking for a similiarily intuitive tool to do the same for rendering JSON schema; but most had custom syntax, requiring to duplicate the `description` for each property.

@dadav [helm-schema](https://github.com/dadav/helm-schema) does exactly that, integrating with already present _helm-docs_ comments, while being able to specify additional schema related properties for automatically creating `values.schema.json` from `values.yaml`.


benjagm (approved):
Thanks for creating this great tool! Congrats!

PR #664, opened 2024-04-18
Change format keyword positioning

bhavukkalra:
**What kind of change does this PR introduce?**
PR moves the `format` section of the documentation from the misplaced [string](https://json-schema.org/understanding-json-schema/reference/string) page to under the more generic [type](https://json-schema.org/understanding-json-schema/reference/type) as discussed in [Issue #759](https://github.com/json-schema-org/json-schema-spec/issues/759). Which aims to remove the Misinterpretation that the reader might get that `format` as a keyword is only for string types which is not the case.

Also, Add further clarification that `format` as a keyword is generic. Using these lines added further

image

**Issue Number:**
-  Closes #187 

**Screenshots**

- Added section under [type](https://json-schema.org/understanding-json-schema/reference/type) page

image

- removed from `format` section from [string](https://json-schema.org/understanding-json-schema/reference/string) page

image

**Summary**
PR aims to re-adjust the positioning of `format` keyword definitions. To remove any confusion of it being a `string` specific keyword as reported by the users here [Slack Archives](https://json-schema.slack.com/archives/C5CF75URH/p1562596499404200)

**Does this PR introduce a breaking change?**
No

CC - @benjagm @gregsdennis @jviotti

PR #963, opened 2024-09-21
Docs: Improve format documentation

### LINKED ISSUE #187: Format is not only allowed for string values
At https://json-schema.org/understanding-json-schema/reference/string.html#format it says:
> The format keyword allows for basic semantic validation on certain kinds of **string** values that are commonly used. 

As discussed in issue https://github.com/json-schema-org/json-schema-spec/issues/759, format is also defined for non-string types. Therefore, the description should be made more general and moved to another place as it doesn't really fit where it is now. I guess it's better placed at the end of https://json-schema.org/understanding-json-schema/reference/type.html or so...

nikhilkalburgi:

Resolves the Docs error by relocating the Format section and also incorporating the suggested changes from the participants of PR #664 

**Issue Number:**

-  Closes #187
-  Related to #664

**Summary**

I have implemented the changes that were discussed in PR #664. Additionally, I have also checked the grammatical errors. Sorry! I did not ask to assign it to me as it was pending for the last few months. 

If you have any other suggestions, let me know.

No


benjagm:
Can you please modify the description of the PR to make it more self-explicative? Thanks a lot!

DhairyaMajmudar (approved):
Looks good 🚀

pages/understanding-json-schema/reference/type.md
```diff
@@ -175,3 +175,64 @@ types. For example, numeric types have a way of specifying a numeric
...
+
+Optionally, validator [implementations](../../learn/glossary#implementation) can provide a configuration option to enable `format` to function as an assertion rather than just an annotation. That means that validation fails when, for example, a value with a `date` format isn't in a form that can be parsed as a date. This allows values to be constrained beyond what other tools in JSON Schema, including [Regular Expressions](../../understanding-json-schema/reference/regular_expressions), can do.
+
+> Implementations may provide validation for only a subset of the built-in formats or do partial validation for a given format. For example, some implementations may consider a string an email if it contains an `@`, while others might perform additional checks for other aspects of a well-formed email address.
+
+The JSON Schema specification has a bias toward networking-related formats, likely due to its roots in web technologies. However, custom formats may also be used if the parties exchanging the JSON documents share information about the custom format types. A JSON Schema validator will ignore any format type it does not understand.
```
gregsdennis, line 189: > The JSON Schema specification has a bias toward networking-related formats, likely due to its roots in web technologies.

I don't know how important this really is.  I think it's sufficient to say that the spec defines some but you can define your own if all consumers agree what that custom format means (and they all configure their validators to behave the same).

pages/understanding-json-schema/reference/type.md
```diff
@@ -175,3 +175,64 @@ types. For example, numeric types have a way of specifying a numeric
...
+
+### Built-in Formats
+
+It should be noted that `format` is not limited because it only defines a specific set of valid values for formats. Users may define their own custom keywords to work with any specific data type, such as `integer`, `double`, `float`, etc. Below, we cover the commonly used `string` formats specified in the JSON Schema specification.
+
+#### Dates and Times
+
+Dates and times are represented in [RFC 3339, section 5.6](https://tools.ietf.org/html/rfc3339#section-5.6). This is a subset of the date format also commonly known as [ISO8601 format](https://www.iso.org/iso-8601-date-and-time-format.html).
```
gregsdennis, line 197: > This is a subset of the date format also commonly known as ISO8601 format.

This is not true.  Here's a really good comparison: https://ijmacd.github.io/rfc3339-iso8601/

gregsdennis (changes requested):

pages/understanding-json-schema/reference/type.md
nikhilkalburgi: Hi @gregsdennis 

**Original:**

The JSON Schema specification has a bias toward networking-related formats, likely due to its roots in web technologies. However, custom formats may also be used if the parties exchanging the JSON documents share information about the custom format types. A JSON Schema validator will ignore any format type it does not understand.

**Revised:**

The JSON Schema specification defines some standard formats, but you can also define custom formats as long as all consumers agree on their meaning and configure their validators to handle them consistently. A JSON Schema validator will ignore any format type it does not understand.

Do you agree with this?

pages/understanding-json-schema/reference/type.md
nikhilkalburgi: Hi @gregsdennis ,

I have two different suggestions for this:

**1:**
Dates and Times
Dates and times are represented in [RFC 3339, section 5.6](https://tools.ietf.org/html/rfc3339#section-5.6). While RFC 3339 is often considered a subset of the [ISO 8601 format](https://www.iso.org/iso-8601-date-and-time-format.html), there are important differences between the two. For a detailed comparison, refer to this resource: [RFC 3339 vs. ISO 8601](https://ijmacd.github.io/rfc3339-iso8601/).


**2:**
Dates and times are represented in [RFC 3339, section 5.6](https://tools.ietf.org/html/rfc3339#section-5.6). While RFC 3339 is often seen as a simpler version of the [ISO 8601 format](https://www.iso.org/iso-8601-date-and-time-format.html), there are some differences. RFC 3339 focuses on a smaller set of date and time formats and has stricter rules. It doesn't include all the features of ISO 8601, like week dates or very long years. For a clear comparison, check this resource: [RFC 3339 vs. ISO 8601](https://ijmacd.github.io/rfc3339-iso8601/).

What do you think?

gregsdennis: Is 3339 considered a simpler version of 8601?  I never thought that. I always thought they were just different. I was actually surprised when I discovered how much they overlap.

nikhilkalburgi: RFC 3339 is a simplified and restricted subset of ISO 8601.

They have some similarities like Both formats support the basic date-time structure using the format `YYYY-MM-DDTHH:MM:SS` and some differences like ISO 8601 allows reduced precision in the time format (e.g, only specifying the hour: 14:30) and RFC 3339 requires the full `HH:MM:SS` format.

ISO 8601 provides more options than RFC 3339.

So, what should be our conclusion for this? Do you prefer one out of the above two options or you have anything else to add?

gregsdennis: > RFC 3339 is a simplified and restricted subset of ISO 8601.

I have a problem with the word "subset".  It's not a subset.  There are 3339 formats that 8601 doesn't accept.  In the comparison, you can plainly see this.  It's not a subset because the 3339 circle did not exist completely within the 8601 circle.

nikhilkalburgi: Then, I think a better way to express this is:

Dates and times are represented in [RFC 3339, section 5.6](https://tools.ietf.org/html/rfc3339#section-5.6). While RFC 3339 shares many similarities with the date format also commonly known as [ISO8601 format](https://www.iso.org/iso-8601-date-and-time-format.html), they are distinct formats. RFC 3339 follows stricter rules and includes some formats that ISO 8601 does not accept. At the same time, RFC 3339 leaves out certain features of ISO 8601, such as week dates or very long years. For a clear comparison, refer to [RFC 3339 vs. ISO 8601](https://ijmacd.github.io/rfc3339-iso8601/).

Is it better now? Anything else to be included?

gregsdennis: Now you have the opposite problem that "distinct" means there's no overlap.  Just say they're separate specs.  The RFC is by IETF, and the ISO is by, well, ISO.  They're different because they are attempts to standardize dates by two different standardization bodies.

nikhilkalburgi: Dates and times are represented in RFC 3339, section 5.6. RFC 3339 and ISO 8601 are standards from different standardization bodies. RFC 3339 is from the IETF and ISO 8601 from the ISO. They are separate date format specifications with overlapping features. RFC 3339 enforces stricter rules and supports certain formats that ISO 8601 does not, while omitting features like week dates and very long years, which are part of ISO 8601. To explore their similarities and differences, refer to: RFC 3339 vs. ISO 8601.
--------------------------------------------------------
I have rephrased it as per your suggestion. I will add the link to IETF, ISO official site and RFC vs ISO.

gregsdennis: I think we're getting lost in the weeds here.  We should **focus on RFC3339** and _maybe_ casually remark that, while it overlaps somewhat with ISO8601, it is a different specification with a different set of formats.

mwadams: Why are we mentioning the ISO standard at all? It isn't actually relevant. There are innumerable other standards that it *isn't* defined by.

gregsdennis: That's really what I'm trying to get at.

benjagm:
I am adding @jagpreetrahi to the discussion as they have been working on this in other [PR](https://github.com/json-schema-org/website/pull/975). Now all the conversation is here.

pages/understanding-json-schema/reference/type.md
gregsdennis: Still waiting on a resolution here.  I'm happy removing any mention of ISO8601.

nikhilkalburgi:
Hi @gregsdennis ,

If we remove the mention of ISO8601 then it becomes:

Dates and times are represented in [RFC 3339, section 5.6](https://tools.ietf.org/html/rfc3339#section-5.6). RFC 3339 is a specification from the Internet Engineering Task Force (IETF) that enforces rules for date and time formats. It defines a precise format for representing dates and times, ensuring consistency across systems and protocols.

Please add to it if anything else comes to your mind.

gregsdennis:
Just leave it at the first sentence

benjagm:
Hi @nikhilkalburgi we are very close to have this ready. Can you please take a look to the last comment?

nikhilkalburgi:
Hi @benjagm , @gregsdennis , I have updated the PR with the changes suggested in the last comment. Can you please verify it?

image

pages/understanding-json-schema/reference/type.md
```diff
@@ -175,3 +175,64 @@ types. For example, numeric types have a way of specifying a numeric
...
+It should be noted that `format` is not limited because it only defines a specific set of valid values for formats. Users may define their own custom keywords to work with any specific data type, such as `integer`, `double`, `float`, etc. Below, we cover the commonly used `string` formats specified in the JSON Schema specification.
+
+#### Dates and Times
+
+Dates and times are represented in [RFC 3339, section 5.6](https://tools.ietf.org/html/rfc3339#section-5.6). RFC 3339 is a specification from the Internet Engineering Task Force (IETF).
+
+- `"date-time"`: Date and time together, for example, `2018-11-13T20:20:39+00:00`.
+- `"time"`: <StarInline label="New in draft 7" /> Time, for example, `20:20:39+00:00`.
+- `"date"`: <StarInline label="New in draft 7" /> Date, for example, `2018-11-13`.
+- `"duration"`: <StarInline label="New in draft 2019-09" /> A duration as defined by the [ISO 8601 ABNF for "duration"](https://datatracker.ietf.org/doc/html/rfc3339#appendix-A). For example, `P3D` expresses a duration of 3 days.
```
gregsdennis, line 202: We link to RFC3339 here, but say that it's from ISO8601.  I think the text is good, but maybe we should link to 8601 directly rather than to the 3339 appendix. 

The text in the spec is:

> The duration format is from the ISO 8601 ABNF as given in Appendix A of RFC 3339.

@mwadams thoughts?

gregsdennis (approved):
I think this is a lot better. Thanks. Simple and to the point. 

Just one other question.  Looking for more opinions.

pages/understanding-json-schema/reference/type.md
jdesrosiers: Doesn't ISO keep their standards behind a paywall? Linking to ISO8601 may not be an option.

gregsdennis: Oh, yeah.  That's dumb.  I guess that's why we copied everything relevant into that appendix.

pages/understanding-json-schema/reference/type.md
```diff
@@ -175,3 +175,64 @@ types. For example, numeric types have a way of specifying a numeric
...
+
+> Implementations may provide validation for only a subset of the built-in formats or do partial validation for a given format. For example, some implementations may consider a string an email if it contains an `@`, while others might perform additional checks for other aspects of a well-formed email address.
+
+The JSON Schema specification has a bias toward networking-related formats, likely due to its roots in web technologies. However, custom formats may also be used if the parties exchanging the JSON documents share information about the custom format types. A JSON Schema validator will ignore any format type it does not understand.
+
+### Built-in Formats
+
+It should be noted that `format` is not limited because it only defines a specific set of valid values for formats. Users may define their own custom keywords to work with any specific data type, such as `integer`, `double`, `float`, etc. Below, we cover the commonly used `string` formats specified in the JSON Schema specification.
```
jdesrosiers, line 193: Formats can only apply to JSON data types which doesn't include `integer`, `double`, and `float`. Although any type can be used, the only types that would make any sense to use with `format` are `string` and `number`. I think the important thing this paragraph should be communicating is that although the JSON Schema specification only defines `string` formats, it's allowed to define `number` formats as well.

Here's my suggestion, but it could probably use a little more work.

```suggestion
It should be noted that `format` is not limited to a specific set of valid values or types. Users may define their own custom keywords including ones that work with JSON data types other than `string`, such as `number`. Below, we cover the formats specified in the JSON Schema specification.
```

jdesrosiers (changes requested):
This mostly looks great. There's just one part that needs one more revision for accuracy.

harshita9104:
Hello @jdesrosiers ,

I have gone through the changes you have requested can i go ahead and make the changes so that this issue is fully solved.

Thanking you in anticipation.

nikhilkalburgi:
Hi @benjagm , @gregsdennis , I have updated the PR with the suggestions from @jdesrosiers.

gregsdennis (approved):

jdesrosiers (approved):

benjagm:
Very happy about merging this one. Thanks everyone!!

PR #998, opened 2024-10-03
Fix hyper-schema page links

jdesrosiers:
I noticed that the links on the hyper-schema page point to the raw xml on github. These links should point to the compiled and published documents. This PR fixes those links.


jdesrosiers:
Something strange is happening with these links. They go to the homepage, but then when I refresh they go to the right place. So the paths are correct. This feels like a bug in Next.js, but maybe it's something related to the markdown-to-html conversion?

benjagm (approved):
Thanks for all the fixes in the content of this page!

PR #1007, opened 2024-10-07
Fix hyperschema links again

jdesrosiers:
This fixes the problem with the links I mentioned in https://github.com/json-schema-org/website/pull/998#issuecomment-2392505313. I figured out that it works when the links use `..`. These links should be equivalent indicating that something is wrong with the way the way the website is setup or a bug in Next.js. This works around the problem, but doesn't fix the underlying issue. It will bite us again if not addressed in the future.


benjagm (approved):
Thanks Jason!

Issue #999, opened 2024-10-04
🐛 Bug:  @cfworker/json-schema is not showing the bowtie badges in the details modal window


benjagm:
### Describe the bug

`@cfworker/json-schema` is not showing the bowtie badges in the details modal window

<img width="877" alt="Screenshot 2024-10-04 at 09 06 48" src="https://github.com/user-attachments/assets/45e1fabe-0b95-4b8f-b6bd-f7491a8d79b4">

### Steps To Reproduce

Go to tooling page and click in cfworker/json-schema to show the details.

### Expected Behavior

Show the bowtie fields.

### Device Information [optional]

```markdown
- OS:
- Browser:
- version:
```

### Are you working on this issue?

No

### Do you think this work might require an [Architectural Decision Record (ADR)]? (significant or noteworthy)

No

RonakSurana-2001:
I am not able to understand can you explain more clearly

Mohith1490:
Hey @benjagm according to above details  "js-json-schema" should look like a badge, like above one right.

DarhkVoyd:
Hi everyone,

Thanks for your interest in this issue! I wanted to let you know that I’m currently working on resolving this. Since this has become a bit more complex than just adding badges and it is no longer a Hacktoberfest issue, I'll be handling the fix. I'll update the label to reflect this.

Please feel free to reach out if you have any questions or need clarification.

PR #1009, opened 2024-10-09
[tools]: Fix tools bowtie outlink and badges integration

DarhkVoyd:
This PR aims to fix the `/tools` bowtie outlink and badges integration. Previously, Bowtie IDs were manually maintained in the data model, which was error-prone and also required updates to tooling data for new/updated bowtie identifiers. This has been fixed with:

- **Removed manual Bowtie ID maintenance** from the data and data model, fully transitioning to using the Bowtie API for dynamic matching.
- Implemented a new matching approach:
  1. We first attempt an **exact match** between the tool’s source URL and the Bowtie report.
  2. If an exact match isn’t found, we perform a **root URL match**, checking if the tool's URL matches the root structure (e.g., `https://<git-provider>.com/<org>/<project>`).
  
Let's test this approach to see how well it works. If needed, there is another potential solution where we match nested paths and pop off segments until we reach the root path, providing more flexibility in matching. We can consider switching to this approach if the current root path matching runs into any limitations.

Closes: #999


benjagm (approved):
Seems to work! Much better this way. Thanks!!

PR #1157, opened 2024-12-10
Add apideck as sponsor

benjagm:
**Summary**

This PR adds Apideck as sponsor as they started to pay again the bronze tier.


Relequestual (approved):
Thanks for picking up on this and making the PR!

PR #1206, opened 2024-12-23
docs: add note clarifying multipleOf implementation

### LINKED ISSUE #1113: [📝 Docs]: Erroneous example for `multipleOf`

The spec of json-schema says that a number `x` satisfies a `multipleOf: m` schema when the result of `x / m` is an integer/ However, in most programming languages, the floating-point numbers are conforming to the ieee754 standard, more precisely, to its binary floating-point specification. In that case, 4.02 is not a multiple of 0.01 (as can be seen in python where `float(402) == 4.02 / 0.01` returns `False`).

The example should be updated to be valid in most programming languages and specify that floating-point multiples are not what they appear to be. For example, 4.2 is a multple of 0.05 according to the spec and when doing the computation in binary floating-points.

### Code of Conduct

- [X] I agree to follow this project's Code of Conduct

inclinedadarsh:

Documentation update

**Issue Number:**

-  Closes #1113 

**Screenshots/videos:**

image

**Summary**

Add a note explaining the difference between float-point implementation of programming languages and json schema

No, it does not.

**Other**

I believe this note can be clearer, better if I also provide an example, not sure about it though. Please let me know about it.

### DIFF
--- a/pages/understanding-json-schema/reference/numeric.md
+++ b/pages/understanding-json-schema/reference/numeric.md
@@ -162,7 +162,7 @@ The multiple can be a floating point number:
 4.021
 ```
 
-<Infobox label="Note"> The JSON Specification allows an arbitrary implementation of the `multipleOf` keyword and is not constrained by the floating-point behavior of most programming languages. In contrast, most programming languages adhere to the IEEE754 standard for floating-point numbers, which can result in discrepancies between validation according to the JSON Specification and the behavior in programming languages. </Infobox>
+<Infobox label="Note"> The JSON Schema Specification defines numerical precision independently of the IEEE 754 standard. This means developers do not need to worry about the typical limitations of floating-point arithmetic, such as precision loss or representation errors, which are common in most programming languages. </Infobox>
 
 <Keywords label="single: number; range single: maximum single: exclusiveMaximum single: minimum single: exclusiveMinimum" />
 
--- a/pages/understanding-json-schema/reference/numeric.md
+++ b/pages/understanding-json-schema/reference/numeric.md
@@ -162,7 +162,7 @@ The multiple can be a floating point number:
 4.021
 ```
 
-<Infobox label="Note"> The JSON Schema Specification defines numerical precision independently of the IEEE 754 standard. This means developers do not need to worry about the typical limitations of floating-point arithmetic, such as precision loss or representation errors, which are common in most programming languages. </Infobox>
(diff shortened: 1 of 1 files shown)


jdesrosiers (changes requested):
Thanks for giving this a start. There are a few things that need a second draft.

> The JSON Specification allows an arbitrary implementation of the `multipleOf` keyword

The JSON _Schema_ Specification defines `multipleOf`, not the JSON Specification. There is nothing arbitrary about `multipleOf`. It has a precise definition and consistent behavior across implementations. I'm not sure what this statement is trying to communicate to the reader, but it's full of inaccuracies.

> and is not constrained by the floating-point behavior of most programming languages.

This part is correct.

> In contrast, most programming languages adhere to the IEEE754 standard for floating-point numbers, which can result in discrepancies between validation according to the JSON Specification and the behavior in programming languages.

This is correct, but not the way I would frame it. I would say that with JSON Schema you don't need to worry about the limitations of IEEE754 floating-point numbers like you do in most programming languages.

inclinedadarsh:
Thanks a lot for the review @jdesrosiers 

Thank you specifying exact changes needed, I'll be fixing these and would push the changes as soon as possible.

pages/understanding-json-schema/reference/numeric.md
```diff
@@ -162,6 +162,7 @@ The multiple can be a floating point number:
 4.021
 ```
 
+<Infobox label="Note"> The JSON Schema Specification defines numerical precision independently of the IEEE 754 standard. This means developers do not need to worry about the typical limitations of floating-point arithmetic, such as precision loss or representation errors, which are common in most programming languages. </Infobox>
```
jdesrosiers, line 165: You still have this a little confused. The JSON specification defines numbers. The JSON Schema specification defines `multipleOf` using JSON numbers.

jdesrosiers (changes requested):
This sounds better but there's still an inaccuracy.

pages/understanding-json-schema/reference/numeric.md
inclinedadarsh: Hey, so sorry for this.
Can you redirect me to a page where I can read more about it, to clear my confusion?

jdesrosiers: It's the [JSON specification](https://www.rfc-editor.org/rfc/rfc8259#section-6) that defines numbers, not the JSON Schema specification. That's the only thing that needs to change. The JSON Schema specification also [mentions](https://json-schema.org/draft/2020-12/draft-bhutton-json-schema-01#section-4.2.1) the JSON data model, but just as a reader convenience.

jdesrosiers (approved):

PR #1219, opened 2024-12-27
Add/comment/multipleof

akshat09867:
# Documentation Update: multipleOf Validation Specification

Related to #1113

## Summary
Updates multipleOf validation specification to address IEEE 754 floating-point precision issues, providing clearer guidance on implementation.

## Documentation Changes
- Added warning about floating-point precision limitations
- Clarified validation requirements
- Added practical implementation guidance
- Updated examples to be more IEEE 754 friendly

### DIFF
--- a/public/draft/2020-12/draft-bhutton-json-schema-validation-01.html
+++ b/public/draft/2020-12/draft-bhutton-json-schema-validation-01.html
@@ -1745,12 +1745,9 @@ li > p:last-of-type {
                         A numeric instance is valid only if division by this keyword's value results in
                         an integer.<a href="#section-6.2.1-2" class="pilcrow">¶</a></p>
 <p id="section-6.2.1-2">
-                        Limitations:Using a floating-point value for 'multipleOf' may lead to inconsistent validation results due to binary precision issues.
-                         Consider using scaled integers instead.<a href="#section-6.2.1-2" class="pilcrow">¶</a></p>
-<p id="section-6.2.1-2">
-                       Solution: Avoid using floating-point numbers with 'multipleOf' due to potential precision issues. For example, use scaled integers (e.g., represent $4.02 as 402 and 'multipleOf' as 1).
-                       If using floating point then use Floating-point-friendly numbers (like 0.05) as it will reduce the chance of 
-                       precision issues but are still less reliable than integers.<a href="#section-6.2.1-2" class="pilcrow">¶</a></p>
+                        To ensure consistent validation:
+                        - Use scaled integers (e.g., represent $4.02 as 402 with 'multipleOf: 1')
+                        - If floating-point values are necessary, use binary-friendly numbers (e.g., 0.5, 0.25, 0.125)</p><a href="#section-6.2.1-2" class="pilcrow">¶</a></p>
 </section>
 <section id="section-6.2.2">
           <h4 id="name-maximum">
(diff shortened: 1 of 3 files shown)


public/draft/2020-12/draft-bhutton-json-schema-validation-01.html
```diff
@@ -1744,6 +1744,10 @@ <h4 id="name-multipleof">
 <p id="section-6.2.1-2">
                         A numeric instance is valid only if division by this keyword's value results in
                         an integer.<a href="#section-6.2.1-2" class="pilcrow">¶</a></p>
+<p id="section-6.2.1-2">
+                        To ensure consistent validation:
+                        - Use scaled integers (e.g., represent $4.02 as 402 with 'multipleOf: 1')
+                        - If floating-point values are necessary, use binary-friendly numbers (e.g., 0.5, 0.25, 0.125)</p><a href="#section-6.2.1-2" class="pilcrow">¶</a></p>
```
gregsdennis, line 1750: Is this really the recommendation we want to make?

mwadams: I'd say "no". Not least because it isn't really meaningful; why would we be recommending that people only use particular values. And then there is the problem of representation on all platforms (e.g. 1.3, 2.4, 5.6 cannot be represented precisely as a 32 bit float)

gregsdennis: A use case could be defining currency, for example dollars & cents.  You'd want to ensure that your data only has a penny ($0.01) resolution.

But the point is that we don't want to force users into a scenario where they should be trying to figure out the nuances of binary float arithmetic while writing their schemas.

I think that the very most, we should be leaving a note that some implementations may use IEEE7 math.

I do think the way that the spec is worded is pretty cleverly, though.

> 
> [6.2.1.](https://json-schema.org/draft/2020-12/json-schema-validation#section-6.2.1) multipleOf
> 
> The value of "multipleOf" MUST be a number, strictly greater than 0.
> A numeric instance is valid only if division by this keyword's value results in an integer.

In C#, 

```c#
double a = .0075;  // IEEE754
double b = .0001;
decimal c = .0075m;
decimal d = .0001m;

Console.WriteLine("double");
Console.WriteLine(a / b);
Console.WriteLine(a % b);
Console.WriteLine();
Console.WriteLine("decimal");
Console.WriteLine(c / d);
Console.WriteLine(c % d);
```

outputs

```
double
75
9.999999999999937E-05

decimal
75
0.0000
```

So following the method prescribed by the spec (using division instead of modulation) actually produces the correct result, even with IEEE754 math.

inclinedadarsh:
Hello @gregsdennis 

This issue was originally assigned to me & I have also opened a PR #1206 on this. I believe you might have missed that, so please check that out. Also, this PR edits the draft itself, however in a conversation with @valeriahhdez, she told me the following:
> For general context, we don't edit the [draft documents](https://json-schema.org/draft/2020-12/json-schema-validation); we only edit the [documentation pages on the website](https://json-schema.org/docs).

So I don't think it would be meaningful to merge this PR.

---

Moreover, @akshat09867 I understand your excitement to contribute, but please first check if the issue is assigned to someone else. This was mentioned by multiple maintainers in their slack channel recently, so you might want to consider joining that. 

If you don't know where to check if the issue was assigned to someone else, in the issue page, there's a tab on the right side and you'll generally find the **Assignees** the first thing there.
image

jdesrosiers:
@inclinedadarsh is right. I'm closing this PR.

public/draft/2020-12/draft-bhutton-json-schema-validation-01.html
karenetheridge: Storing monetary values as cents is a good idea, to keep all math as integers, but a better way of specifying that in the schema is with `type: integer` not `multipleOf: 1`.

I don't think we want to add anything like `scale` into the spec, though -- this should be defined and controlled by the individual application.

We may want to add a note that tooling might implement the `multipleOf` keyword using modulo calculations, as a lead-in to the warning about imprecision with floating point storage.

PR #1245, opened 2025-01-05
Updated and optimised Dockerfile and modified dockerignore

### LINKED ISSUE #1196: ✨ Enhancement: Update on Dockerfile Implementation for Windows Users
### Is your feature request related to a problem? Please describe

The current implementation of our Dockerfile, particularly for Windows users.

Key Issues:
1. Incompatibility with make Command: Currently, Windows users are unable to utilize the make command within the Docker container. This limitation arises from the differences in command-line tools available on Windows compared to Unix-based systems. As a result, any build processes that rely on make will not function as intended on Windows.

2. Yarn Version Mismatch Error: Additionally, when executing standard Docker commands such as docker build and docker run, users may encounter a yarn version mismatch error. This issue occurs because the version of Yarn installed in the Docker container may differ from the version expected by the application. This discrepancy can lead to build failures and unexpected behavior during runtime.

### Describe the solution you'd like

To added the corepack enable command in the Dockerfile. This adjustment ensures that the correct version of Yarn is used consistently across different environments.

### Describe alternatives you've considered

_No response_

### Additional context

_No response_

### Are you working on this?

Yes

abhayymishraa:
- This PR will fix the bug of Dockerfile and introduce an optimised image
- This PR will also implement the env logic

**Issue Number:**

-  Closes #1196 
-  Related to #1196  
-  Others? 

**Screenshots/videos:**

-  Screenshot from 2025-01-05 14-49-31
-  Screenshot from 2025-01-05 15-01-22

  N/A

**Summary**
 The Dockerfile has been optimized to include proper Yarn 4.4.0 installation from source, enhanced cleanup processes to 
 reduce image size, improved environment file handling, and explicit command execution, resulting in a more efficient and 
 reliable container build process.

 N/A

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -27,12 +27,6 @@ RUN apk add --update git && \
     apk del git curl && \
     rm -rf /var/cache/apk/*
 
-RUN if [ ! -f .env ]; then \
-    cp .env.example .env && rm -rf .env.example; \
-    else \
-    rm -rf .env.example; \
-    fi
-
 EXPOSE 3000
 
 CMD ["yarn", "dev"]
--- a/Dockerfile
+++ b/Dockerfile
@@ -2,16 +2,37 @@ FROM node:20-alpine
 
 WORKDIR /app
 
-COPY package*.json ./
+RUN apk add --no-cache git curl bash && \
+    rm -f /usr/local/bin/yarn /usr/local/bin/yarnpkg && \
+    curl -o /tmp/yarn-4.4.0.tar.gz -L https://github.com/yarnpkg/berry/archive/refs/tags/@yarnpkg/cli/4.4.0.tar.gz && \
+    mkdir -p /opt/yarn && \
+    tar -xzf /tmp/yarn-4.4.0.tar.gz -C /opt/yarn --strip-components=1 && \
+    ln -s /opt/yarn/packages/yarnpkg-cli/bin/yarn.js /usr/local/bin/yarn && \
+    ln -s /opt/yarn/packages/yarnpkg-cli/bin/yarn.js /usr/local/bin/yarnpkg && \
+    rm -rf /tmp/yarn-4.4.0.tar.gz
+
+COPY . .
 
 RUN apk add --update git && \
     git init && \
     git submodule init && \
     git submodule update && \
-    yarn
+    yarn install && \
+    yarn cache clean && \
+    rm -rf .git && \
+    rm -rf /root/.cache && \
+    rm -rf /root/.npm && \
+    rm -rf /root/.yarn && \
+    rm -rf /tmp/* && \
+    apk del git curl && \
+    rm -rf /var/cache/apk/*
 
-COPY . .
+RUN if [ ! -f .env ]; then \
+    cp .env.example .env && rm -rf .env.example; \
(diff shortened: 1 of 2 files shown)


Dockerfile
```diff
@@ -2,16 +2,37 @@ FROM node:20-alpine
...
+    rm -rf /tmp/* && \
+    apk del git curl && \
+    rm -rf /var/cache/apk/*
 
-COPY . .
+RUN if [ ! -f .env ]; then \
+    cp .env.example .env && rm -rf .env.example; \
+    else \
+    rm -rf .env.example; \
+    fi
```
DhairyaMajmudar, line 34: I don't think we need to check if the container has `.env` file or not since there's no use setting environment variables for installing the codebase. You can remove this code and changes from `.dockerignore` also.

abhayymishraa: Sure

DhairyaMajmudar:
Thank you @abhayymishraa for the PR. One questions.... Dockerfile will become compatible with Windows machines also?

Dockerfile
DhairyaMajmudar: No need to open another PR, you can commit to this exiting pr only.

abhayymishraa:
@DhairyaMajmudar  i ran a test in my friend's system (windows 11 )  .  it worked smoothly

Dockerfile
abhayymishraa: yes yes i will commit it

abhayymishraa:
@DhairyaMajmudar  done ✔️

jdesrosiers (changes requested):
This is unmaintainable. If this is the best way to install yarn, I think it's too soon to move away from corepack. Corepack isn't gone yet and it appears that Yarn doesn't have a reasonable path to installing without it yet. I say we wait until this situation improves before trying to remove corepack.

abhayymishraa:
yes , The curl is the way to get that specific package without any problem although i have a alternative i found out recently looking forward to soon make a pr with the changes

Dockerfile
```diff
@@ -2,16 +2,31 @@ FROM node:20-alpine
 
 WORKDIR /app
 
-COPY package*.json ./
```
DhairyaMajmudar, line 5: Don't remove this step, for docker optimisation we use multi copying technique,  where we first copy dependency files (This creates a layer in image) and run the command for installing them (this creates another layer). Whenever changes are made to codebase which doesn't include changes with dependency files the first layer is cached and reused. 

In our case it will be `COPY package.json yarn.lock`

DhairyaMajmudar (changes requested):

Dockerfile
abhayymishraa: Actually if we do that the docker file will stop working i don't know what happens but if i install their then i get this error node modules not found 
```
docker run -it -p 3000:3000 json-schema 
Usage Error: Couldn't find the node_modules state file - running an install might help (findPackageLocation)

$ yarn run [--inspect] [--inspect-brk] [-T,--top-level] [-B,--binaries-only] [--require #0] <scriptName> ...

```

benjagm:
Closed due to lack of activity in the last month.

PR #1357, opened 2025-01-22
Change cypress-ajv-schema-validator from Javascript to JavaScript

Kiina:
Currently on the page the cypress-ajv-schema-validator is the only one that writes it Java**s**cript instead of Java**S**cript. Which means it ends up with it's own entry on:

https://json-schema.org/tools?query=&sortBy=name&sortOrder=ascending&groupBy=languages&licenses=&languages=&drafts=&toolingTypes=&environments=&showObsolete=false#javascript while also breaking the anchor link (since that one is 100% case insensitive).


DhairyaMajmudar (approved):
Thanks for the fix

LGTM 👍🏻

benjagm (approved):
LGTM

PR #1310, opened 2025-01-14
Fix : pro help sidebar alignment and add buttons

Karan-Palan:
**What kind of change does this PR introduce?**

1.  Pro Help Page Sidebar and Navigation Buttons:
- Added "Go Back" and "Up Next" buttons to the Pro Help page for consistent 
- Aligned the sidebar to match the logical flow of pages: Case Studies → FAQ → Pro Help → Similar 
- Fixed layout misalignment issues on the top and left of the Pro Help page.
2. Arrow Navigation Updates Across Pages

**Issue Number:**

-  Closes #1291 

**Screenshots/videos:**

[Screencast from 2025-01-14 13-31-31.webm](https://github.com/user-attachments/assets/0676c54c-a336-4c44-8229-111ea2072204)

**If relevant, did you update the documentation?**
No

**Summary**

#1291  - Fixes Pro Help page sidebar alignment, adds navigation buttons, and standardizes arrow navigation across pages.

**Does this PR introduce a breaking change?**

No

PR #1376, opened 2025-01-25
fix: improve styling in pro-help page

### LINKED ISSUE #1368: Issues with the Pro-Help page
The https://json-schema.org/pro-help page was recently added. While adding my bio, I noticed several issues now that we have more than one bio on the page.

- Inconsistent sizing of avatars -- Juan's is rendering bigger that the others for some reason, which also makes the layout between the cards not line up.
- No padding between the listings -- A little spacing would look nicer.
- The bio should use the markdown component -- Plain text is too limiting.

This screenshot comes from running the site locally with the additional entries.
Image

Thund3rHawk:
Adjust the styling of the pro-help page to address the inconsistent image sizes, wrap the bio inside a markdown component for increasing word limitations and resolve the spacing issues between the cards.

**Issue Number:**

-  Closes #1368 

**Screenshots/videos:**
Screenshot 2025-01-26 at 11 57 41 AM

No

**Summary**
This pr resolves the issue https://github.com/json-schema-org/website/issues/1368


jdesrosiers (changes requested):
The website uses markdown in lots of places already. You shouldn't need to include a new dependency, just use what's already being used on other pages. If you don't find it I'm sure someone can chime in and point you at the right component.

Thund3rHawk:
Sure. Please help.

DarhkVoyd:
@Thund3rHawk, have you had a chance to explore the [StyledMarkdown](https://github.com/json-schema-org/website/blob/main/components/StyledMarkdown.tsx) component that we’ve integrated across our website?

Thund3rHawk:
Hello @DarhkVoyd, thank you for pointing this out! I haven’t had the chance to explore the StyledMarkdown component yet, but I’m looking forward to diving into it. Could you let me know the best place to ask questions if I get stuck along the way?

Thund3rHawk:
@jdesrosiers done. Please check.

DhairyaMajmudar (changes requested):
@Thund3rHawk your PR is failing CI checks pls. fix them by running the command `yarn format:fix` in console.

jdesrosiers (changes requested):
Something's not right. See how Julian an my bios get rendered differently. The line spacing is different. Maybe other things as well, but that's the obvious one.

image

Thund3rHawk:
Hello @jdesrosiers, I’ve noticed an issue with the input item, specifically with the line breaking in the JSON file. Thank you for bringing this to attention, and I’m currently working on resolving it.

jdesrosiers:
> I’ve noticed an issue with the input item, which seems to occur because we are rendering JSON inside a Markdown component. [...] I’m considering creating a JSON-to-Markdown converter.

That doesn't make sense. It shouldn't make any difference where the data came from. It could be JSON, YAML, or even plain text. By the time the component gets it, it's just a string. So, a new component doesn't make sense to me.

> Could you suggest a solution to resolve this issue?

From what I can see, it looks like the `StyledMarkdown` component renders markdown differently if there are newlines present in the string. You get a `span` if it's just one block of text and `p`s and `div`s if there are multiple blocks. That sounds like a bug in the way the `StyledMarkdown` component works. It should be styled consistently in either case. Either it needs to be styled properly for the single block case, or it needs to somehow be forced to always use the multi-block rendering mode. The latter seems like a better approach to me because it's always going to be consistent and there aren't two different places it needs to be styled that can get out of sync.

But, I'm not very familiar with the website's architecture so, @DarhkVoyd or @DhairyaMajmudar can probably give better guidance in this case.

Thund3rHawk:
> From what I can see, it looks like the `StyledMarkdown` component renders markdown differently if there are newlines present in the string.

I think this should be a new issue? And I will work on that.

DhairyaMajmudar:
Hi @jdesrosiers @Thund3rHawk, I explored the root cause of the issue and its not related to `StyledMarkdown` component. 

In the contractors.json file the bio data of Jason consists of special characters `\n\n` which changes its the text type from HTML to markdown as of which such problems are made.

This can be resolved if we replace that special character to `<br/>` HTML tag.

jdesrosiers:
> In the contractors.json file the bio data of Jason consists of special characters \n\n which changes its the text type from HTML to markdown as of which such problems are made.

The bio field should interpreted as markdown, not HTML. It sounds like your solution is backwards. The problem isn't that my bio is being interpreted as markdown. The problem is that the other bios aren't being interpreted as markdown. The bio field should always be interpreted as markdown, not HTML.

Thund3rHawk:
Would it be better if the bio came inside a markdown format instead of JSON? I feel it might be a more efficient solution.

jdesrosiers:
> Would it be better if the bio came inside a markdown format instead of JSON?

That sentence doesn't make sense to me. What does it mean for the bio to come inside a markdown format? Markdown is just a string that uses markdown syntax. The bio a string that uses markdown syntax. What are suggesting should be different?

Thund3rHawk:
I believe the spacing issue might be occurring due to the usage of '\n' within the JSON. From my perspective, adopting Markdown could be a better approach since it eliminates the need for '\n' or any other line-breaking tags.

Thund3rHawk:
Hello @jdesrosiers @DhairyaMajmudar, Please check now the bio renders correctly as expected.

DhairyaMajmudar:
Thank you @jdesrosiers for suggesting the fix, after applying the `forceBlock:true` the contractors profile will look like this 


| Jason's Profile | Jaun's Profile | Julian's Profile |
|---------------------|-------------------|----------------------|
|image|image|image|


We can achieve this by passing a default false prop in `StyledMarkdown` component which can enable or disable `forceBlock` option. By means of this the other places pf website where markdown is used won't be affected from `forceBlock` option and here where we need that it can be set true and taken  in used.

What are your views in this ?

jdesrosiers:
> We can achieve this by passing a default false prop in `StyledMarkdown` component which can enable or disable `forceBlock` option. By means of this the other places of website where markdown is used won't be affected by `forceBlock` option and here where we need that it can be set to true and taken in use.

I think you're assuming that block mode would only be used on this page. That's not correct. The entire site renders markdown in block mode. I'm pretty certain that this page is the only place were the markdown we use is simple enough that markdown-to-jsx decided to render it in inline mode instead of block mode. Everywhere else uses block mode. So, setting `forceBlock: true` should be a no-op for the rest of the site. It tells markdown-to-jsx to choose block mode even though it was already choosing block mode anyway. If there are places on the site where the markdown is getting rendered in inline mode, it's almost certainly just as wrong there as it is on this page and this change would fix those problems too.

So, my solution would be to set `forceBlock: true` and not bother providing a prop to override it. I don't think we ever actually want inline mode and this makes sure this problem doesn't show up in other places in the future. If it turns out that we do need to render in inline mode in the future, we can add an option to force inline mode, but I wouldn't do that until we determine that there's actually a need for it.

@DhairyaMajmudar, I think your solution is ok but I don't think it's the best approach. Your solution is best to make sure we don't accidentally break something now. My solution is best to ensure we don't see this same bug in the future on other pages. Your approach is safer now, but the risk is negligible and has no real consequence, so I'd rather take that risk and protect ourselves from encountering this bug again.

The question remains, are we ok with the style changes introduced by using markdown? Does the rest of the page need additional adjustments to match the markdown styling (and the rest of the site)?

DhairyaMajmudar:
I got your point Jason and it's pretty valid to ensure block mode on all of the website pages where markdown is rendered.

> The question remains, are we ok with the style changes introduced by using markdown? Does the rest of the page need additional adjustments to match the markdown styling (and the rest of the site)?

Yeah there are some changes required to match the overall page with the other pages of the website, like...

- The sidebar component is currently missing which will be resolved after merging of PR: #1310
- There's also a scope of improvements for font sizes and how proof of work links are shown as after applying the suggested fix those links will look a bit more highlighted as compared to the bio section.

jdesrosiers:
@DhairyaMajmudar Sounds good to me!

> There's also a scope of improvements for font sizes and how proof of work links are shown as after applying the suggested fix those links will look a bit more highlighted as compared to the bio section.

I think at a minimum we'll want to update the font color and line spacing to match the markdown. @Thund3rHawk If you implement the fix we've been discussing and at least those minimal style updates, I promise I won't ask for anything else :laughing:.

Thund3rHawk:
@jdesrosiers Absolutely! I’d be happy to implement the fix you are discussed. If you could help implement the fix along with the font color and line spacing updates, I’d be truly grateful.

jdesrosiers (approved):
Thanks to everyone who helped with this!

jdesrosiers:
I squashed commits and rebased to fix merge conflicts. I'm not sure what went wrong exactly, but the changes have been successfully merged. Sorry for any confusion.

PR #1473, opened 2025-03-06
Tools Page Table mobile Adjustments 

### LINKED ISSUE #1442: 🐛 Bug: Tools Page Tables Require Adjusting
### Describe the bug

On the Tools page, the tables are not optimized. Users need to scroll sideways to view all the information, which affects readability and usability mainly on mobile.  

### Steps To Reproduce

1. Go to the Tools page on a mobile device.  
2. Observe that the table content extends beyond the screen width.  
3. Scroll sideways to access the hidden information.  

### Expected Behavior

- Tables should be fully readable without excessive horizontal scrolling.  
- Content should be responsive and properly formatted for mobile screens.  

### Screenshots

this is the current behavior on mobile: 

Image

and for desktop (the bowtie column is being cut off)

Image

and this is my suggestion of how it should look like in mobile:

Image

for desktop it will be a much simpler fix.
 

```markdown
- OS:
- Browser:
- version:
```

Yes

No

idanidan29:

This PR will adjust the tables on the tools page for mobile and improve the desktop table to prevent it from being cut off, ensuring users don't need to scroll sideways to view the full length of the table.

**Issue Number:**

-  Closes #1442

**Screenshots/videos:**

Image
image

**Summary**
I think this is a necessary change, as it will improve the readability of the tools on the page.

no


DhairyaMajmudar (changes requested):
@idanidan29 we're loosing table headers with this new design, is there any solution for that?

image

image

idanidan29:
@DhairyaMajmudar Yes, that's the intended design. It was also suggested in the issue, as keeping the headers would make it too dense, small, and hard to read. But we could add an example card with all the fields labeled by their names at the start, it might work better.  i would like to hear your opinion on this

DhairyaMajmudar:
> @DhairyaMajmudar Yes, that's the intended design. It was also suggested in the issue, as keeping the headers would make it too dense, small, and hard to read. But we could add an example card with all the fields labeled by their names at the start, it might work better. i would like to hear your opinion on this

Remvoing table header won't be a good idea as it will be difficult for user to know what fields are related to which tag, the changes which you're suggesting is also not so optimal as it will add and empty/dummy card. Pls. suggest some more ideas.

benjagm:
Hi @DarhkVoyd can you please provide feedback? There are some good proposals there.

Utkarsh-123github:
@idanidan29 @DhairyaMajmudar @benjagm   I have an idea regarding the design. We can make a two column grid which just shows the tool name and a "Click here" button on each tool  and when user taps then a popover comes which contains all the information along with the links to bowtie and other related links for that particular tool.
What's your take on this ?

Utkarsh-123github:
By implementing two column grid the design will look less cluttered and will have good optimization of space.

idanidan29:
Hey @DhairyaMajmudar, thanks for getting involved! I actually suggested a similar idea to what you brought up in my recent adjustment, but I think a two-column layout might provide too little information about each tool. A card-like approach might be better since adding a "press for more details" option would also take up space. We already have the functionality to show full details when a tool is pressed, so this approach would keep things more efficient. We could, however, leave some info out of the card and make it visible only when pressed, like the Bowtie, for example, as it takes up a whole line of space without using most of it.

idanidan29:
I think we should gather more feedback on that. I’ll post it in the #website thread on Slack and update this PR accordingly.

DarhkVoyd:
> Hi @DarhkVoyd can you please provide feedback? There are some good proposals there.

Sure, is this still in work?

idanidan29:
@DarhkVoyd yes its on https://github.com/json-schema-org/website/pull/1584.

PR #1501, opened 2025-03-11
fix: removed conflicting css property

### LINKED ISSUE #1417: [Code Bug] : Conflicting Tailwind Classes flex and block Used Simultaneously
In the SideBar.tsx component, both flex and block classes are used together at one place, which is incorrect. These classes conflict with each other as they define different display properties. Using both causes issues and results in a runtime error.

Expected Behavior
Only one of the classes should be used depending on the intended layout (either flex or block).

Steps to Reproduce
1. Open the SideBar.tsx component.
2. Observe the usage of both flex and block in the same element.

Adi-204:
Removed conflicting css property from Sidebar.tsx component.

**Issue Number:**
-  Closes #1417 

**Screenshots/videos:**

Below is screenshot after resolving the conflicting CSS.

image

No

No

### DIFF
--- a/components/Sidebar.tsx
+++ b/components/Sidebar.tsx
@@ -65,7 +65,7 @@ const DocLinkBlank = ({
   return (
     <Link
       href={uri}
-      className={classnames('flex text-sm block  py-1 pl-2', {
+      className={classnames('flex text-sm py-1 pl-2', {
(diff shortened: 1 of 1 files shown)


DhairyaMajmudar (approved):
LGTM 👍🏻

PR #1516, opened 2025-03-14
update validate-tooling-data for eliminate case insensitive languages

### LINKED ISSUE #1443: 🐛 Bug: Two JavaScript labels for filter on Tools page
### Describe the bug

On the tools page, on the left side, there is a dropdown for 'Language.' In that dropdown, there are two labels pointing to the same thing: 'JavaScript' and 'Javascript.' The only difference is the capitalization of 's' and 'S.

### Steps To Reproduce

1. Go to tools page(https://json-schema.org/tools).
2. Click on 'Language' dropdown.
3. Scroll down to the names starting from letter 'J'.
4. There you can see 'Javascript' and 'Javascript'.

### Expected Behavior

There should be only one label 'JavaScript'.

### Screenshots

Image

```markdown
- OS: Ubantu 24.04
- Browser: Firefox
- version: 135.0.1 (64-bit)
```

Yes

Yes

vishvboda:
Feature - Adds case-insensitive unique validation for language entries

**Issue Number:**

-  Closes #1443  
-  Related to #___ 
-  Others? 

**Screenshots/videos:**
Forcefully made mistakes in the name of language,
image

Validator finds the mistake,
image

**Summary**
This PR introduces case-insensitive unique validation for language entries in the tooling data to solve several existing problems:
1. Inconsistent language casing across tools (e.g., "JavaScript" vs "javascript" vs "JAVASCRIPT")
2. Potential confusion for users seeing the same language listed multiple times

My solution:
Implements a custom AJV keyword `caseInsensitiveUnique` that:
- Detects and reports case-insensitive duplicates using set
- Provides clear error messages for easy fixes
```
           ajv.addKeyword({
              keyword: 'caseInsensitiveUnique',
              type: 'array',
              validate: function (schema, data) {
                if (!Array.isArray(data)) return false;
                
                const languagesSet = new Set();
                const languagesLowercaseSet = new Set();
                data.forEach((tool) => {
                  if (tool.languages) {
                    tool.languages.forEach((language) => {
                      languagesSet.add(language);
                      languagesLowercaseSet.add(language.toLowerCase());
                    });
                  }
                });
                if (languagesSet.size !== languagesLowercaseSet.size) {
                  console.error('Duplicate languages found');
                  const lowercaseMap = new Map();
                  languagesSet.forEach((language) => {
                    lowercaseMap.set(
                      language.toLowerCase(), 
                      (lowercaseMap.get(language.toLowerCase()) || 0) + 1
                    );
                  });
                  
                  lowercaseMap.forEach((value, key) => {
                    if (value > 1) {
                      console.log('Duplicate found for:', key);
                    }
                  });
                  validate.errors = [{
                    keyword: 'caseInsensitiveUnique',
                    message: 'array contains case-insensitive duplicates',
                    params: { keyword: 'caseInsensitiveUnique' }
                  }];
                  return false;
                }
                return true;
              }
            });
```
 

✅ **Yes**

**Impact:**  
This PR enforces **case-insensitive uniqueness** for language entries. Any existing tooling data that includes language names with inconsistent casing—such as `"JavaScript"` and `"javascript"`—will now **fail validation**. This change helps eliminate redundancy and confusion caused by duplicate entries with different letter cases.

**Who is affected:**  
Tool maintainers and contributors who have added language entries with varying casing.

**Migration Path:**  
Update your `languages` arrays to ensure that each language appears only once in a consistent format, preferably matching the casing defined in the schema enum. For example:

```yaml
# ❌ Before
languages:
  - "JavaScript"
  - "javascript"
  - "Go"
  - "go"

# ✅ After
languages:
  - "JavaScript"
  - "Go"

### DIFF
--- a/.github/workflows/validate-tooling-data.yml
+++ b/.github/workflows/validate-tooling-data.yml
@@ -47,6 +47,49 @@ jobs:
             const data = yaml.load(fs.readFileSync(dataPath, 'utf-8'));
             const ajv = new Ajv({ allErrors: true });
             addFormats(ajv);
+
+            ajv.addKeyword({
+              keyword: 'caseInsensitiveUnique',
+              type: 'array',
+              validate: function (schema, data) {
+                if (!Array.isArray(data)) return false;
+                
+                const languagesSet = new Set();
+                const languagesLowercaseSet = new Set();
+                data.forEach((tool) => {
+                  if (tool.languages) {
+                    tool.languages.forEach((language) => {
+                      languagesSet.add(language);
+                      languagesLowercaseSet.add(language.toLowerCase());
+                    });
+                  }
+                });
+                if (languagesSet.size !== languagesLowercaseSet.size) {
+                  console.error('Duplicate languages found');
+                  const lowercaseMap = new Map();
+                  languagesSet.forEach((language) => {
+                    lowercaseMap.set(
+                      language.toLowerCase(), 
+                      (lowercaseMap.get(language.toLowerCase()) || 0) + 1
+                    );
+                  });
+                  
+                  lowercaseMap.forEach((value, key) => {
+                    if (value > 1) {
+                      console.log('Duplicate found for:', key);
+                    }
+                  });
+                  validate.errors = [{
+                    keyword: 'caseInsensitiveUnique',
+                    message: 'array contains case-insensitive duplicates',
+                    params: { keyword: 'caseInsensitiveUnique' }
+                  }];
+                  return false;
+                }
+                return true;
+              }
+            });
+
             const validate = ajv.compile(schema);
             const valid = validate(data);
             if (!valid) {
--- a/data/tooling-data.schema.json
+++ b/data/tooling-data.schema.json
@@ -88,6 +88,7 @@
       "languages": {
         "description": "The language or languages a tool is built in. In the case of a validator, this will likely be the language it is written in. In the case of a conversion or transformation tool, these are the languages that are supported in some capacity.",
         "type": "array",
+        "caseInsensitiveUnique": true,
         "items": {
           "description": "Individual language name, from the list unless not included.",
           "type": "string",


jagpreetrahi:
Hey @Vishv04, Can you tell me what change you made for this file 'validate-tooling-data.yml'.  Is it necessary to change something there because the reason for failing this PR is changing this unauthorized file

Vishv04:
> Hey @Vishv04, Can you tell me what change you made for this file 'validate-tooling-data.yml'. Is it necessary to change something there because the reason for failing this PR is changing this unauthorized file

Hi @jagpreetrahi, I added a custom caseInsensitiveUnique rule in validate-tooling-data.yml to detect case-insensitive duplicates in the languages array—like treating "JavaScript" and "javascript" as the same. It logs duplicates and throws a validation error if found.

benjagm:
@jviotti can i get your support to help me reviewing this solution?

jviotti:
I don't think there is any reason to add a new keyword here. Why not just use `pattern` to set a regular expression that only allows lowercase strings?

vishvboda:
Thank you @benjagm @jviotti for your response on this PR. 

> I don't think there is any reason to add a new keyword here. Why not just use `pattern` to set a regular expression that only allows lowercase strings?

Yes, using a pattern works, but it feels a bit like we're controlling user input, since users will naturally write "Javascript" instead of "javascript" (just my assumption, it might be wrong). I created the custom keyword to avoid forcing users to change how they enter data. Still, using a pattern could also be a workable option.

jviotti:
I really suggest just using `pattern` to avoid the extra complexity of a new keyword (mainly with AJV). The convention can be to just force everybody to write languages in lowercase.

vishvboda:
Okay @jviotti, I will use pattern and fix this issue. Thank you for your help.

Tushar8466:
Nice update 👍

However, I noticed the CI check for unauthorized file changes is failing. Could you please review and fix this so the PR can pass all checks?

PR #1579, opened 2025-03-22
Fix: Make entire button area clickable by applying styles directly to <Link> 

### LINKED ISSUE #1562: Buttons are working only when we click on text inside
Button link are working only when we click on the text.

**Current Video**

https://github.com/user-attachments/assets/154c774a-8e12-4434-bf89-eab04c6fe227

techmannih:
 Make entire button area clickable by applying styles directly to <a> tag

**Issue Number:**

-  Closes #1562

### DIFF
--- a/components/Layout.tsx
+++ b/components/Layout.tsx
@@ -343,9 +343,9 @@ const Footer = () => (
           </a>
         </div>
         <div className='flex flex-col text-center sm:text-left'>
-          <a href='/overview/code-of-conduct' className='text-white mb-2'>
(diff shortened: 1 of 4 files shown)


DarhkVoyd (changes requested):
Since we're using Next.js page routes, I’d also suggest replacing <a> with Next.js <Link> where applicable.

DarhkVoyd (changes requested):
Since we're using Next.js page routes, I’d also suggest replacing <a> with Next.js <Link> where applicable.

DarhkVoyd (approved):
LGTM!

PR #1954, opened 2025-11-23
docs: correct type list and fix typos in type-specific keywords page #1952

### LINKED ISSUE #1952: [📝 Docs]: fixs typo's in the 'reference/type.md'

This documentation page contains several inaccuracies and formatting problems that reduce clarity and correctness.

### Code of Conduct

- [x] I agree to follow this project's Code of Conduct

DivyanshuVortex:

Documentation fix (typo corrections and type-list accuracy improvements)

**Issue Number:**

- Closes #1952

**Screenshots/videos:**

N/A (text-only documentation changes)

Yes — this PR directly updates the documentation page.

**Summary**

This PR corrects multiple inaccuracies in the *Type-specific Keywords* documentation:

- Fixed the primitive type list and ensured accuracy with the JSON Schema specification.
- Cleaned several typos, spacing issues, and minor wording inconsistencies.
- Ensured examples refer only to valid JSON Schema types.

These improvements increase correctness, clarity, and consistency across the documentation.

No. Documentation-only changes.

# Checklist

### DIFF
--- a/pages/understanding-json-schema/reference/type.md
+++ b/pages/understanding-json-schema/reference/type.md
@@ -16,14 +16,16 @@ At its core, JSON Schema defines the following basic types:
 - [array](../../understanding-json-schema/reference/array)
 - [boolean](../../understanding-json-schema/reference/boolean)
 - [null](../../understanding-json-schema/reference/null)
-- [numeric types](../../understanding-json-schema/reference/numeric)
+- [number](../../understanding-json-schema/reference/numeric)
 - [object](../../understanding-json-schema/reference/object)
-- [regular expressions](../../understanding-json-schema/reference/regular_expressions)
 - [string](../../understanding-json-schema/reference/string)
 
 
 These types have analogs in most programming languages, though they may
 go by different names.
+
+> Note: JSON does not define an `integer` type. In JSON Schema, `"integer"` is shorthand for `"number"` with `"multipleOf": 1`, used to represent whole-number numeric values.
+
 
 [tabs-start "Language-specific info"]
 
@@ -127,7 +129,7 @@ to decode ``Int`` from a non-integer number in JSON.
 
 The `type` keyword can take two forms:
 
-1. **A single string**. When it is a single string, it must be one of the types mentioned above (`array`, `boolean`, `integer`, `number`, `null`, `object`, `regular expressions`, or `string`). This specifies that the instance data is only valid when it matches that specific type. 
+1. **A single string**. When it is a single string, it must be one of the types mentioned above (`array`, `boolean` , `number`, `null`, `object` or `string`). This specifies that the instance data is only valid when it matches that specific type. 
 
 Here is an example of using `"number"` as a single string value:
 
@@ -176,7 +178,7 @@ JSON Schema offers a variety of keywords to validate data against specific types
 | Type Keyword | Specific Keywords | Description |
 |---|---|---|
 | [array](../../understanding-json-schema/reference/array)        | `items`, `additionalItems`, `minItems`, `maxItems`, `uniqueItems` | Define item schemas, additional item handling, item count constraints, and uniqueness. |
-| [number](../../understanding-json-schema/reference/numeric)       | `minimum`, `maximum`, `exclusiveMinimum`, `exclusiveMaximum`, `multipleOf` | Define numeric ranges, including exclusive bounds and divisibility. |
+| [number](../../understanding-json-schema/reference/numeric) | `minimum`, `maximum`, `exclusiveMinimum`, `exclusiveMaximum`, `multipleOf` | Define numeric ranges and allowed increments. |
 | [object](../../understanding-json-schema/reference/object)      | `required`, `properties`, `additionalProperties`, `patternProperties`, `minProperties`, `maxProperties`, `dependencies` | Define required properties, property schemas, additional property handling, pattern-based property matching, and property count constraints. |
 | [string](../../understanding-json-schema/reference/string)       | `minLength`, `maxLength`, `pattern`, `format`| Restrict string length, pattern matching, and format validation (e.g., email, date). |
 
@@ -206,7 +208,7 @@ The JSON Schema specification has a bias toward networking-related formats due t
 
 ### Built-in Formats
 
-It should be noted that `format` is not limited to a specific set of valid values or types. Users may define their own custom keywords including ones that work with JSON data types other than `string`, such as `number`. Below, we cover the formats specified in the JSON Schema specification.
+It should be noted that `format` is not limited to a specific set of valid values or types. Users may define additional custom formats, but the built-in formats apply only to the `string` type. Below, we cover the formats defined in the JSON Schema specification.
 
 #### Dates and Times
 
--- a/pages/understanding-json-schema/reference/type.md
+++ b/pages/understanding-json-schema/reference/type.md
@@ -72,7 +72,7 @@ floating-point.
 [tab "Perl"]
 
 The following table maps from the names of JSON types to
-their analogous types in Objective-C:
+their analogous types in Perl:
 
 | JSON    | Perl                                             |
 |---------|--------------------------------------------------|
@@ -85,7 +85,7 @@ their analogous types in Objective-C:
 
 [tab "Objective-C"]
 
-The following table maps from the names of JavaScript types to
+The following table maps from the names of JSON types to
 their analogous types in Objective-C:
 
 | JSON    | Objective-C  |
@@ -105,7 +105,7 @@ value for identity (pointer equality) to `@YES` and `@NO` constants.
 
 [tab "Swift"]
 
-The following table maps from the names of JavaScript types to
+The following table maps from the names of JSON types to
 their analogous types in Swift:
 
 | JSON    | Swift      |
@@ -129,7 +129,7 @@ The `type` keyword can take two forms:
 
 1. **A single string**. When it is a single string, it must be one of the types mentioned above (`array`, `boolean`, `integer`, `number`, `null`, `object`, `regular expressions`, or `string`). This specifies that the instance data is only valid when it matches that specific type. 
 
-Here is an example of using the `string` keyword as a single string:
+Here is an example of using the `number` keyword as a single string:
 
 ```json
 // props { "isSchema": true }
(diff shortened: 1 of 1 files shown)


pages/understanding-json-schema/reference/type.md
```diff
@@ -129,7 +129,7 @@ The `type` keyword can take two forms:
 
 1. **A single string**. When it is a single string, it must be one of the types mentioned above (`array`, `boolean`, `integer`, `number`, `null`, `object`, `regular expressions`, or `string`). This specifies that the instance data is only valid when it matches that specific type. 
 
-Here is an example of using the `string` keyword as a single string:
+Here is an example of using the `number` keyword as a single string:
```
jdesrosiers: We usually only refer to the property names as "keywords", not their values. So, `type` is a keyword and `"number"` is the value of that keyword.

```suggestion
Here is an example of using `"number"` as a single string value:
```

jdesrosiers (changes requested):
Thanks for fix this. There's one more thing we can fix while we're here. See the inline comment.

pages/understanding-json-schema/reference/type.md
DivyanshuVortex: Hey @jdesrosiers , pushed the requested changes. Lmk if you need tweaks .

pages/understanding-json-schema/reference/type.md
jdesrosiers: "numeric types" is not a type.


```suggestion
- [integer](../../understanding-json-schema/reference/numeric)
- [number](../../understanding-json-schema/reference/numeric)
```

jdesrosiers: "regular expressions" is not a type.


```suggestion
```

pages/understanding-json-schema/reference/type.md
```diff
@@ -129,7 +129,7 @@ The `type` keyword can take two forms:
 
 1. **A single string**. When it is a single string, it must be one of the types mentioned above (`array`, `boolean`, `integer`, `number`, `null`, `object`, `regular expressions`, or `string`). This specifies that the instance data is only valid when it matches that specific type.
```
jdesrosiers: "regular expressions" is not a type.

```suggestion
1. **A single string**. When it is a single string, it must be one of the types mentioned above (`array`, `boolean`, `integer`, `number`, `null`, `object`, or `string`). This specifies that the instance data is only valid when it matches that specific type. 
```

jdesrosiers (commented):
Sorry, I noticed a couple more things that are wrong. Not sure when this got so broken, but I appreciate you fixing these issues as well. Thanks.

pages/understanding-json-schema/reference/type.md
```diff
@@ -47,8 +48,7 @@ analogous types in Python:
 analogous to `unicode` on Python 2.x and `str` on
 Python 3.x.
 
-[2] JSON does not have separate types for integer and
-floating-point.
+[2] JSON defines a single numeric literal, but JSON Schema distinguishes between `number` (any numeric value) and `integer` (no fractional part).
```
jdesrosiers: This is not correct. JSON Schema is uses the mathematical value, not the way number is represented. So, `1.0` is considered an integer even though it contains a fractional part.

I think you can revert this and the other footnote changes. The original was fine.

pages/understanding-json-schema/reference/type.md
```diff
@@ -178,6 +181,7 @@ JSON Schema offers a variety of keywords to validate data against specific types
 | [array](../../understanding-json-schema/reference/array)        | `items`, `additionalItems`, `minItems`, `maxItems`, `uniqueItems` | Define item schemas, additional item handling, item count constraints, and uniqueness. |
 | [number](../../understanding-json-schema/reference/numeric)       | `minimum`, `maximum`, `exclusiveMinimum`, `exclusiveMaximum`, `multipleOf` | Define numeric ranges, including exclusive bounds and divisibility. |
 | [object](../../understanding-json-schema/reference/object)      | `required`, `properties`, `additionalProperties`, `patternProperties`, `minProperties`, `maxProperties`, `dependencies` | Define required properties, property schemas, additional property handling, pattern-based property matching, and property count constraints. |
+| [integer](../../understanding-json-schema/reference/numeric)     | Same as `number` | Integers use numeric constraints but must not contain a fractional part. |
```
jdesrosiers: Again, the fractional part is not relevant. I don't think we need this row, but maybe we can update the `number` row to `number/integer` or something like that.

pages/understanding-json-schema/reference/type.md
```diff
@@ -206,7 +210,7 @@ The JSON Schema specification has a bias toward networking-related formats due t
 
 ### Built-in Formats
 
-It should be noted that `format` is not limited to a specific set of valid values or types. Users may define their own custom keywords including ones that work with JSON data types other than `string`, such as `number`. Below, we cover the formats specified in the JSON Schema specification.
+It should be noted that `format` is not limited to a specific set of valid values or types. Users may define custom formats through vocabularies, but the built-in formats apply only to strings.
```
jdesrosiers: I think this was more clear before except for the error of calling it "custom keywords" instead of "custom formats". Also, formats can't be added through vocabularies.

jdesrosiers (changes requested):
Thanks for continuing to iterate on this.

There's a nuance about types that isn't communicated here properly. (Again, I don't know how this got so bad. It didn't used to be like this.) JSON has six types: null, boolean, number, string, array, and object. "Integer" _is not_ a JSON type.

JSON Schema does something confusing where it allows for a seventh value for `type` called "integer". This _does not_ mean that JSON or JSON Schema supports integers. It's effectively a shorthand for `{ "type": "number", "multipleOf": 1 }` for describing numbers that are restricted to the set of integers.

So, adding "integer" to all the tables is not the right thing to do.

pages/understanding-json-schema/reference/type.md
DivyanshuVortex: Thanks for the clarification. I misunderstood the role of "integer" in JSON Schema and treated it as a base JSON type instead of a specialized constraint on "number".

pages/understanding-json-schema/reference/type.md
```diff
@@ -127,9 +129,9 @@ to decode ``Int`` from a non-integer number in JSON.
 
 The `type` keyword can take two forms:
 
-1. **A single string**. When it is a single string, it must be one of the types mentioned above (`array`, `boolean`, `integer`, `number`, `null`, `object`, `regular expressions`, or `string`). This specifies that the instance data is only valid when it matches that specific type. 
+1. **A single string**. When it is a single string, it must be one of the types mentioned above (`array`, `boolean` , `number`, `null`, `object` or `string`). This specifies that the instance data is only valid when it matches that specific type.
```
jdesrosiers:
```suggestion
1. **A single string**. When it is a single string, it must be one of the types mentioned above (`array`, `boolean`, `number`, `null`, `object` or `string`). This specifies that the instance data is only valid when it matches that specific type. 
```

pages/understanding-json-schema/reference/type.md
```diff
@@ -206,7 +208,7 @@ The JSON Schema specification has a bias toward networking-related formats due t
 
 ### Built-in Formats
 
-It should be noted that `format` is not limited to a specific set of valid values or types. Users may define their own custom keywords including ones that work with JSON data types other than `string`, such as `number`. Below, we cover the formats specified in the JSON Schema specification.
+It should be noted that `format` is not limited to a specific set of valid values or types. Users may define additional custom formats, but the built-in formats apply only to the `string` type. Below, we cover the formats defined in the JSON Schema specification.
```
jdesrosiers: This still doesn't quite communicate what it needs to say. It's supposed to clarify that just because all of the formats in the spec apply to strings, doesn't mean custom formats can't apply to types other than strings. I'm not understanding that from the current wording. The original wording was fine other than calling them "custom keywords" instead of "custom formats".

pages/understanding-json-schema/reference/type.md
```diff
@@ -16,15 +16,17 @@ At its core, JSON Schema defines the following basic types:
...
-- [regular expressions](../../understanding-json-schema/reference/regular_expressions)
 - [string](../../understanding-json-schema/reference/string)
 
 
 These types have analogs in most programming languages, though they may
 go by different names.
 
+> Note: JSON does not define an `integer` type. In JSON Schema, `"integer"` is shorthand for `"number"` with `"multipleOf": 1`, used to represent whole-number numeric values.
+
+
```
jdesrosiers: At no point do we document that the `type` keyword takes `"integer"`. I think something like this works better.

```suggestion
The `type` keyword defines the following types:

- [array](../../understanding-json-schema/reference/array)
- [boolean](../../understanding-json-schema/reference/boolean)
- [null](../../understanding-json-schema/reference/null)
- [integer](../../understanding-json-schema/reference/numeric)
- [number](../../understanding-json-schema/reference/numeric)
- [object](../../understanding-json-schema/reference/object)
- [string](../../understanding-json-schema/reference/string)

> Note: JSON does not define an `integer` type. In JSON Schema, `"type": "integer"` is shorthand for `"type": "number", "multipleOf": 1`, and is used to represent whole-number numeric values.

These types have analogs in most programming languages, though they may
go by different names.

```

jdesrosiers (changes requested):

pages/understanding-json-schema/reference/type.md
```diff
@@ -127,9 +127,9 @@ to decode ``Int`` from a non-integer number in JSON.
 
 The `type` keyword can take two forms:
 
-1. **A single string**. When it is a single string, it must be one of the types mentioned above (`array`, `boolean`, `integer`, `number`, `null`, `object`, `regular expressions`, or `string`). This specifies that the instance data is only valid when it matches that specific type. 
+1. **A single string**. When it is a single string, it must be one of the types mentioned above (`array`, `boolean`, `number`, `null`, `object` or `string`). This specifies that the instance data is only valid when it matches that specific type.
```
jdesrosiers, line 130: This should include `"integer"` because it's talking about the values the `type` keyword accepts.

jdesrosiers (changes requested):
Sorry, I missed one in the last review.

PR #2031, opened 2025-12-25
fix: use unique stable keys for ToolingTable rows to improve reconcil…

### LINKED ISSUE #2027: 🐛 Bug: React anti-pattern - Array index used as key in ToolingTable
### Describe the bug

In `pages/tools/components/ToolingTable.tsx` (line 188), array index is being used as the key prop for mapped `<tr>` elements. This is a React anti-pattern that can cause rendering issues, incorrect component updates, and bugs when the list is modified (sorted, filtered, or items are added/removed).

Using array indices as keys is problematic because:
- Indices are not stable across re-renders when items are reordered, filtered, or removed
- Can cause React to incorrectly reuse component instances
- May lead to stale state and incorrect UI updates
- Negatively impacts performance and reconciliation

### Steps To Reproduce

1. Navigate to the `/tools` page
2. Open browser DevTools and observe the React component tree
3. Apply filters or sorting to the tools table
4. Observe that components may not update correctly due to index-based keys
5. In the code, check `pages/tools/components/ToolingTable.tsx` line 188
6. Note the use of `index` as the key prop in the `.map()` function

### Expected Behavior

The key prop should use a unique, stable identifier from the tool object itself instead of the array index. 

Recommended fix:
{toolsByGroup[group].map((tool: JSONSchemaTool) => {
  const toolKey = tool.source || tool.name || tool.url;
  return (
    <tr key={toolKey}>  // ✅ Using stable, unique identifier
      ...
    </tr>
  );
})}

This ensures:
- Stable keys that persist across re-renders
- Correct component reconciliation
- Better performance
- Predictable behavior when sorting/filtering

### Screenshots

_No response_

```markdown
- OS:
- Browser:
- version:
```

Yes

No

Shuklax:
Bugfix / Refactoring.

**Issue Number:**
Closes #2027

**Summary** 
This PR fixes the "index as key" anti-pattern in the ToolingTable component. Using indices as keys causes React to incorrectly reuse component instances during filtering and sorting, leading to stale UI states and performance issues.

**My solution** implements a composite stable key using the tool's name and its source/homepage URL. I have applied this fix to both the Desktop and Mobile table views to ensure a consistent experience across all devices.

### DIFF
--- a/pages/tools/components/ToolingTable.tsx
+++ b/pages/tools/components/ToolingTable.tsx
@@ -183,9 +183,10 @@ const ToolingTable = ({
                   if (bowtieData) {
                     tool.bowtie = bowtieData;
                   }
+                  const toolKey = `${tool.name}-${tool.source || tool.homepage || index}`;
                   return (
                     <tr
-                      key={index}
+                      key={toolKey}
                       className='flex w-full hover:bg-gray-100 dark:hover:bg-slate-700 cursor-pointer'
                       onClick={() => openModal(tool)}
                     >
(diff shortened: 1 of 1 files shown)


Utkarsh-123github:
I’m closing this PR for now. For future contributions, please wait until the issue is assigned to you before opening a PR. We want to foster a collaborative environment, not a competitive one, and this helps avoid overlapping efforts.
Thank you!

PR #2051, opened 2025-12-31
fix: replace array index keys with stable identifiers in ToolingTable

### LINKED ISSUE #2027: 🐛 Bug: React anti-pattern - Array index used as key in ToolingTable

Sam-61s:
- This PR fixes a React anti-pattern in `ToolingTable` where the array index was used as the `key` prop for table rows.

Using array indices as keys can cause incorrect component reconciliation, UI inconsistencies, and stale renders when the list is sorted, filtered, or updated. This PR replaces index-based keys with stable, unique identifiers derived from the tool data itself, ensuring predictable behavior and better performance.

**Issue Number:**
- Closes #2027

**Screenshots/videos:**
- Not applicable (no UI changes)

- Not required

- No

# Checklist
- [x] Read, understood, and followed the contributing guidelines

### DIFF
--- a/pages/tools/components/ToolingTable.tsx
+++ b/pages/tools/components/ToolingTable.tsx
@@ -80,18 +80,6 @@ const ToolingTable = ({
 
   const groups = Object.keys(toolsByGroup);
 
-  const generateToolKey = (tool: JSONSchemaTool, group: string): string => {
-    const parts = [group];
-    if (tool.source) parts.push(tool.source);
-    if (tool.name) parts.push(tool.name);
-    if (tool.homepage && tool.homepage !== tool.source)
-      parts.push(tool.homepage);
-    if (tool.license) parts.push(tool.license);
-    if (tool.status) parts.push(tool.status);
-    if (tool.lastUpdated) parts.push(tool.lastUpdated);
-    return parts.join('|');
-  };
-
   const openModal = (tool: JSONSchemaTool) => {
     setSelectedTool(tool);
     postAnalytics({
@@ -195,10 +183,9 @@ const ToolingTable = ({
                   if (bowtieData) {
                     tool.bowtie = bowtieData;
                   }
-                  const toolKey = generateToolKey(tool, group);
                   return (
                     <tr
-                      key={toolKey}
+                      key={`${group}-${tool.name}-${tool.source || tool.homepage || ''}-${tool.toolingTypes?.join(',') || ''}-${tool.status || ''}`}
                       className='flex w-full hover:bg-gray-100 dark:hover:bg-slate-700 cursor-pointer'
                       onClick={() => openModal(tool)}
                     >
@@ -293,10 +280,9 @@ const ToolingTable = ({
                   if (bowtieData) {
                     tool.bowtie = bowtieData;
                   }
-                  const toolKey = generateToolKey(tool, group);
                   return (
                     <tr
-                      key={toolKey}
+                      key={`${group}-${tool.name}-${tool.source || tool.homepage || ''}-${tool.toolingTypes?.join(',') || ''}-${tool.status || ''}`}
                       className='border-b border-gray-200 hover:bg-gray-100 dark:hover:bg-slate-700 cursor-pointer'
                       onClick={() => openModal(tool)}
                     >


Sam-61s:
hey @Utkarsh-123github,
could you please review this PR 
Thanks !!

pages/tools/components/ToolingTable.tsx
```diff
@@ -178,14 +178,15 @@ const ToolingTable = ({
                 </tr>
               </thead>
               <tbody>
-                {toolsByGroup[group].map((tool: JSONSchemaTool, index) => {
+                {toolsByGroup[group].map((tool: JSONSchemaTool) => {
                   const bowtieData = getBowtieData(tool);
                   if (bowtieData) {
                     tool.bowtie = bowtieData;
                   }
+                  const toolKey = tool.source || tool.name;
```
Utkarsh-123github: Are you sure that the toolKey is unique ?
If no, then this might give errors.

Sam-61s: hi @Utkarsh-123github,
I fixed the uniqueness in `ToolingTable` by added a `generateToolKey` fcn to build composite key from (`group name`, `source`, `name`, `homepage`, `licence`, `status` and `lastUpdated`). It handles all edgecases to give stable unique keys.

Utkarsh-123github (changes requested):
To be very honest, this isn't a good fix. We should not add fields like license, lastUpdated etc.
One suggestion from my side is that maybe you can try composite key, something like this:
key={`${tool.name}-${tool.source || tool.homepage }`}

Sam-61s:
> To be very honest, this isn't a good fix. We should not add fields like license, lastUpdated etc. One suggestion from my side is that maybe you can try composite key, something like this: key={`${tool.name}-${tool.source || tool.homepage }`}

if we use `key={${tool.name}-${tool.source || tool.homepage}}`, there will at least one real collision. the SS attached having two separate entries in `tooling-data.yaml` use the same name and source (and no homepage) can differing only by status/lastUpdated .
 That’s why I went with a slightly richer composite key that also includes status/lastUpdated (and the group) to keep those entries unique. if it looks bulky, I can make it a little compact
 `const generateToolKey = (tool: JSONSchemaTool, group: string): string => [
  group,
  tool.source ?? tool.homepage ?? tool.name,
  tool.status ?? '',
  tool.lastUpdated ?? '',
].join('|');`

Screenshot 2026-01-07 153732

Screenshot 2026-01-07 153748

Utkarsh-123github:
> > To be very honest, this isn't a good fix. We should not add fields like license, lastUpdated etc. One suggestion from my side is that maybe you can try composite key, something like this: key={`${tool.name}-${tool.source || tool.homepage }`}
> 
> if we use `key={${tool.name}-${tool.source || tool.homepage}}`, there will at least one real collision. the SS attached having two separate entries in `tooling-data.yaml` use the same name and source (and no homepage) can differing only by status/lastUpdated . That’s why I went with a slightly richer composite key that also includes status/lastUpdated (and the group) to keep those entries unique. if it looks bulky, I can make it a little compact `const generateToolKey = (tool: JSONSchemaTool, group: string): string => [ group, tool.source ?? tool.homepage ?? tool.name, tool.status ?? '', tool.lastUpdated ?? '', ].join('|');`
> 
> Screenshot 2026-01-07 153732 Screenshot 2026-01-07 153748

Ahh, good catch🙌🏻 
But I ran a script to test your new suggested solution, but it was failing for few cases.
Another suggestion is that you can use this :
key={`${tool.name}-${tool.source || tool.homepage || ''}-${tool.toolingTypes?.join(',') || ''}-${tool.status || ''}`}

I checked with the script and this wasn't failing

Sam-61s:
with this, 
 key={`${tool.name}-${tool.source || tool.homepage || ''}-${tool.toolingTypes?.join(',') || ''}-${tool.status || ''}`}
In group validator:
Key = JsonSchema.Net-https://github.com/gregsdennis/json-everything-validator,code-to-schema,schema-to-data,annotations,bundler-
In group bundler:
Key = JsonSchema.Net-https://github.com/gregsdennis/json-everything-validator,code-to-schema,schema-to-data,annotations,bundler-
Same key in both groups.

but if we include group, 
key={`${group}-${tool.name}-${tool.source || tool.homepage || ''}-${tool.toolingTypes?.join(',') || ''}-${tool.status || ''}`} 
this will work perfectly...
It passes the script without failing

Sam-61s:
> with this, key={`${tool.name}-${tool.source || tool.homepage || ''}-${tool.toolingTypes?.join(',') || ''}-${tool.status || ''}`} In group validator: Key = JsonSchema.Net-https://github.com/gregsdennis/json-everything-validator,code-to-schema,schema-to-data,annotations,bundler- In group bundler: Key = JsonSchema.Net-https://github.com/gregsdennis/json-everything-validator,code-to-schema,schema-to-data,annotations,bundler- Same key in both groups.
> 
> but if we include group, key={`${group}-${tool.name}-${tool.source || tool.homepage || ''}-${tool.toolingTypes?.join(',') || ''}-${tool.status || ''}`} this will work perfectly... It passes the script without failing

Hey @Utkarsh-123github , could you please take a look and let me know what you think?

PR #2187, opened 2026-01-31
feat: Add 'Jump to Section' menu to Tooling page

### LINKED ISSUE #2104: ✨ Enhancement: Add "Jump to Section" navigation when Group By is active
### Problem
When users enable "Group By" (e.g., by Tooling Type), the list becomes very long with multiple section headers (Validators, Annotations, etc.). Navigating to a specific category (like "Benchmarks") requires significant scrolling.

Image

### Proposed Solution
Add a "Jump to Section" dropdown or navigation bar next to the Group By controls.

- **Dynamic:** It should only populate with the groups currently visible in the table.
- **Interaction:** Selecting a category should smooth-scroll the page directly to that section header.
- **Benefit:** This improves usability for the "Grouped" view without forcing users to use the sidebar filters to hide content.

I would like to work on this implementation using React refs or IDs for the section headers.

### Describe alternatives you've considered

_No response_

### Additional context

_No response_

### Are you working on this?

Yes

AnujAga2005:
Feature (UX Improvement)

**Issue Number:**
- Closes #2104 

**Screenshots/videos:**
image

**Summary**
This PR introduces a **"Jump to Section"** menu to the Tooling page to improve navigation.

**Motivation:**
The Tooling page contains a long list of resources (Validators, Generators, Utilities, etc.). Scrolling to find a specific category can be tedious. This feature allows users to quickly jump to the section they need.

**Implementation Details:**
- Added a `GroupByMenu` component that dynamically lists available sections based on the current sort/filter.
- Implemented smooth scrolling to the selected section.

No

# Checklist


AnujAga2005:
Hi @AgniveshChaubey , the PR is ready for review, please take a look at this whenever you have time. Thanks!

pages/tools/components/ToolingTable.tsx
```diff
@@ -113,7 +113,10 @@ const ToolingTable = ({
       {groups.map((group) => (
         <section key={group} className='mb-12 text-left'>
           {group !== 'none' && (
-            <div className='mb-10 px-4 w-full bg-gray-100 dark:bg-slate-900'>
+            <div
+              id={group.toLowerCase().replace(/\s+/g, '-')}
+              className='mb-10 px-4 w-full bg-gray-100 dark:bg-slate-900 scroll-mt-24'
```
AgniveshChaubey: I cant see any effect of this class name

pages/tools/index.page.tsx
```diff
@@ -153,12 +153,12 @@ export default function ToolingPage({
         </div>
 
         {/* FIX 1: mx → px, and overflow-x-hidden */}
-        <div className='w-full grid grid-cols-1 lg:grid-cols-4 px-4 md:px-12 min-h-screen overflow-x-hidden'>
+        <div className='grid grid-cols-1 lg:grid-cols-4 mx-4 md:mx-12 min-h-screen'>
```
AgniveshChaubey: what changes do these introduces?

pages/tools/components/GroupByMenu.tsx
```diff
@@ -35,22 +62,86 @@ const GroupByMenu = ({ transform, setTransform }: GroupByMenuProps) => {
...
+              {group.label}
+            </Button>
+          ))}
+        </div>
+
+        {groupedBy !== 'none' && activeSections.length > 0 && (
+          <div className='relative' ref={dropdownRef}>
+            <div className='mx-2 h-6 w-px bg-slate-300 dark:bg-slate-700 inline-block align-middle' />
+
+            <Button
```
AgniveshChaubey, line 101: why do we need a button here? I think simply `select` element would do the work, right? If we wanna maintain consistency, maybe we can modify the `Button` component to return `select` element when some specific prop is passed (like buttonType = 'select')

pages/tools/components/GroupByMenu.tsx
```diff
@@ -1,15 +1,42 @@
...
+  transform,
+  setTransform,
+  activeSections = [],
+}: GroupByMenuProps) => {
   const groupedBy = transform.groupBy;
 
+  const [isOpen, setIsOpen] = useState(false);
+  const dropdownRef = useRef<HTMLDivElement>(null);
+
+  useEffect(() => {
```
AgniveshChaubey, line 27: I think this useEffect would not be required we use `select` element for dropdown

AgniveshChaubey:
Additionally, on smaller screen, UI doesnt look good. We should handle that too
image

pages/tools/components/ToolingTable.tsx
AnujAga2005: It's added to support Direct links. If a user visits the page via a shared URL like .../tools#validator, the browser naturally jumps to that ID on load. Without scroll-mt-24, the sticky navbar covers the section header. This class ensures that direct links land correctly with the header visible. It doesn't affect the button click behavior.

pages/tools/components/GroupByMenu.tsx
AnujAga2005: I used this to match the visual style of the neighboring 'Tooling Type' and 'Language' buttons. A native select usually renders with OS-default styling and might look inconsistent here.

pages/tools/components/GroupByMenu.tsx
AnujAga2005: If we decide to switch to a native <select> (as discussed in the comment above), I will remove this entirely.

pages/tools/index.page.tsx
AnujAga2005: I originally tweaked these (switching px to mx and removing overflow) while debugging. However, after re-testing, I realized that it works perfectly fine with the original classes.

I have reverted these changes.

AnujAga2005:
I've fixed the mobile UI layout (added wrapping) and reverted the overflow and mx changes in index.page.tsx.
Happy to make further updates if needed. Thanks!

pages/tools/components/ToolingTable.tsx
AnujAga2005: I was re-evaluating the implementation and saw that you were right. This wasn't required in our codebase because we already have spacing handled for the navbar.

It was a slight overlook on my side, so I have reverted this change as well. Apologies for the mistake and thanks for the guidance!

AnujAga2005:
Thanks for the detailed review @AgniveshChaubey!

I've pushed updates to address the feedback. Here is a summary of the changes:

Mobile Layout: I fixed the buttons on smaller screens by updating the container to use flex-wrap. The "Jump to Section" button now wraps gracefully to the next line when space is limited.

Reverted Unnecessary CSS:

index.page.tsx: I reverted the mx/px and overflow changes. After re-testing, I confirmed the Sidebar works perfectly with the original layout classes, so I reverted these to minimize the diff.

scroll-mt-24: You were right—this was redundant. I verified that the site's global styles already handle the scroll offset for direct links (headers aren't hidden behind the navbar), so I removed this class.

Custom Dropdown vs. Native <select>:

I currently kept the custom GroupByMenu component to ensure Visual Consistency with the neighboring "Tooling Type" and "Language" buttons (which are custom UI). A native <select> might look visually disconnected here.

However, if you strongly prefer the simplicity of a native select, I am happy to refactor this! (In that case, the useEffect hook would also be removed).

Ready for another look when you have a moment. Thanks again for the guidance!

AnujAga2005:
One question regarding the Sidebar UX: While testing, I noticed that when filter sections are expanded, the "Apply Filters" button gets pushed off-screen, forcing the user to scroll down to find it.

Should I fix that in this PR (making the button sticky/always visible at the bottom), or would you prefer I open a separate Issue/PR for that? 

image

AnujAga2005:
Hi @AgniveshChaubey, I'm closing this PR for now. Due to the recent major updates in the website repo over the last two weeks, several conflicts have emerged in the layout and table logic.

To keep the codebase clean and ensure the UI remains stable, I will open a fresh PR tomorrow with all the features (Jump to Section menu and mobile layout fixes) integrated into the current main branch. Thanks!

AnujAga2005:
@AgniveshChaubey  should i keep the custom dropdown or use native ?

AgniveshChaubey:
I think you can style the **native** dropdown to make it look like the custom one.

AnujAga2005:
@AgniveshChaubey  Okk i will try this approach and one more thing i might take a week to open the new PR because of my tight college schedule and internal exams. Sorry for the delay.

AgniveshChaubey:
That's totally okay! No hurry at all!

PR #2229, opened 2026-02-11
Introduce a trademark policy page

jviotti:
This is a long due task that we, as an independent open source project, should have done ages ago. In summary:

- This project represents the brand "JSON Schema", including more than just the specification, but learning resources, tooling lists, marketing website, blog, and much more
- We are not a legal entity of any sorts, which means properly registering "JSON Schema" as a trademark is harder
- At this point in time, we are not part of any foundation that would deal with this for us. Furthermore, even under our previous IETF involvement, IETF _does not_ own trademarks of its projects, which means this falls on our shoulders even if associated with an institution such as the IETF
- However, the idea of "Common Law Trademarks" offers a reasonable protection and ownership over our project name and does not require registration. Sounds like a no-brainer, and makes a full registration easier in the future if we ever want to do it
- Common Law Trademarks are commonplace in open source projects. The Linux Foundation, for example, actively makes use of them. As a notable example, "OpenAPI" is held by Linux Foundation using common law only
- Because we don't have a single owner of this project nor a legal entity to represent us, I'm proposing we award trademark rights to the members of the Technical Steering Committee at any point of time. That way, any decision around the project name is also subjected to the same governance decision framework we have for other things

There are many resources out there clearly documenting this and how to do it, which I considered to create this pull request. Some links and quotes:

- https://google.github.io/opencasebook/trademarks/#common-law-trademarks
- https://fossmarks.org (mostly https://fossmarks.org/post/5registration/)
- https://www.linuxfoundation.org/blog/blog/open-source-communities-and-trademarks-a-reprise

Here is a longer explanation based on my research today:

A common question when asserting trademark rights over a long-running open source project is whether previous maintainers, authors, or contributors might have competing claims. Under common-law trademark principles, rights belong to whoever is actively using the mark in commerce (and note that commerce does not mean for profit. The Google guide above clarifies this), not to whoever used it first historically. If a party stops using a mark and makes no effort to maintain it, their claim weakens and eventually lapses through what trademark law calls "abandonment." What matters is continuous, active, public use: maintaining the website, publishing specifications, running the community infrastructure, organising the conference, and being the group that developers find and rely on when they look for "JSON Schema". The current active members of the TSC have been doing all of this without interruption.

The JSON Schema project has had various contributors and maintainers over its >10 year history, and not all transitions were clean. Some previous participants became inactive, stopped contributing, and ceased any involvement with the project years ago. This does not create a problem for our claim. If anything, it strengthens it. Common-law trademark rights are not like copyright, where an author retains rights indefinitely regardless of activity. Trademark rights require ongoing use and ongoing stewardship of the mark. The current TSC members are the ones who maintain json-schema.org, publish and develop the specification, run the GitHub organisation, manage the test suite, and operate the community channels. That unbroken chain of active use is what a common-law trademark claim is built on, and it sits with us.

There is also a misconception that because "JSON Schema" has been around for a long time, it can no longer be trademarked. This mixes up two different things. A name can become unprotectable if it becomes "generic", meaning the public uses it to describe an entire category of products rather than a specific source. But that is not what has happened here. When developers say "JSON Schema," they mean this specific project, this specification, and this community.  They are not using the term generically to describe any possible schema language for JSON. The length of time the name has been in use actually works in our favour: it is evidence of long-standing, continuous use, which is one of the strongest foundations for a common-law trademark claim.


pages/overview/trademark-policy/_index.md
```diff
@@ -0,0 +1,9 @@
+## JSON Schema Trademark Policy
+
+"JSON Schema" and the JSON Schema logo are trademarks held by the active members of the Technical Steering Committee (TSC) as per the [JSON Schema governance document](https://github.com/json-schema-org/community/blob/main/GOVERNANCE.md). Trademark stewardship follows active governance as TSC membership changes over time and trademark rights transfer to the current active members rather than remaining with past members. These trademarks are maintained on behalf of the JSON Schema community to ensure that the name remains a reliable indicator of the specification and resources maintained by the project.
```
jdesrosiers, line 3: This is perfect, but it probably means we need to officially document who the TSC is, which I don't think we have at all yet.

jviotti: True, but I wouldn't make it strong requirement. Nobody will care right now that there is not a super explicit list. The only case where we would need to explain this is we would actually engage in a legal debate, which is the worst case. And there, even things like the TSC team on GitHub as a disclosed proof, etc would suffice. We can make the case that's clear even though there is no list right now.

So just to clarify: I don't think we should get lost in the details and only publish something if all the possible pre-requisites are public and well crafted. That doesn't really diminishes our legal stand. Though we definitely should for other reasons.

jdesrosiers: Agreed. It's not a blocker and doesn't need to be part of this PR or happen before this PR. But, it is something that needs to be done regardless.

PR #2300, opened 2026-03-01
fix: correct JSON string boundary detection for escaped backslashes

### LINKED ISSUE #2237: 🐛 Bug: JSON parsing bug: Incorrect handling of backslash-escaped quotes
### Describe the bug

Describe the bug
The custom JSON parser (implemented in lib/getPartsOfJson.ts) incorrectly identifies string boundaries when a string ends with an escaped backslash.

The parser uses the regex /(?<!\)"/g to find the closing quote of a string. This logic is flawed because it assumes any backslash immediately preceding a quote escapes that quote. However, in JSON, a backslash can itself be escaped.

For example, in the valid JSON string "slash: \", the first backslash escapes the second, and the following quote is the closing quote. Because the current regex sees a backslash before the quote, it fails to match it as the end of the string, causing the syntax highlighting and path detection to break for the entire remainder of the code block.

### Steps To Reproduce

1.) Open any markdown file in the project (e.g., pages/understanding-json-schema/basics.md).
2.) Add a JSON code block that includes a trailing escaped backslash:

json

{
  "bad_string": "backslash: \\",
  "next_key": "this will be broken"
}
3.) Run yarn dev and navigate to that page.
4.) Observe: The syntax highlighting "leaks" out of the first string. The comma, the second key, and the rest of the block will all be colored as if they were still part of the first string.

### Expected Behavior

The parser should correctly handle parity for escape sequences. A quote is only escaped if it is preceded by an odd number of backslashes.

### Screenshots

Normal :-

Image

Buggy :-

Image

```markdown
- OS:
- Browser:
- version:
```

Yes

Yes

student-ankitpandit:
Bugfix

**Issue Number:**
Fixes #2237

**Screenshots/videos:**
Before the fix:
image
After the fix:
image

**How should this be tested?**
Open any markdown file in the project (eg: pages/understanding-json-schema/basics.md).
Try adding even number of trailing backslashes in any json code block.

No

# Checklist

- [X] Read, understood, and followed the [contributing guidelines](https://github.com/json-schema-org/website/blob/main/CONTRIBUTING.md).

### DIFF
--- a/lib/getPartsOfJson.ts
+++ b/lib/getPartsOfJson.ts
@@ -9,7 +9,7 @@ const regexNumber = /^\s*(?<number>-?\d+(\.\d+)?([Ee][+-]?\d+)?)\s*$/g;
 const regexString = /^\s*(?<string>"(\\"|[^"])*")\s*$/g;
 const regexBoolean = /^\s*(?<boolean>true|false)\s*$/g;
 const regexNull = /^\s*(?<null>null)\s*$/g;
-const regexDoubleQuote = /(?<!\\)"/g;
+const regexDoubleQuote = /(?<!(?<!\\)(\\\\)*\\)"/g;
 const regexCommaOrEndOfLine = /,/g;
 
 export type SyntaxPart = {


Adityakk9031:
@student-ankitpandit hey close this pr

PR #2407, opened 2026-06-03
Add blog post on making JSON strings work for every language

### LINKED ISSUE #2405: New blog post, Making JSON Strings Work for Every Language

JSON is a very successful data format, but its string type is very simple. A value such as a title and name often needs more than Unicode characters. Consumers may need to know the language of the text and its direction, e.g., when displaying right-to-left scripts such as Arabic or Hebrew. Without that metadata, applications fall back to guessing, and guessing often fails.

The W3C Internationalization Working Group has a document on this topic: [Strings on the Web: Language and Direction Metadata](https://www.w3.org/TR/string-meta/), which is a practical guide for specifications and formats that exchange strings. I'd like to write a blog post about this.

(See also https://github.com/json-schema-org/json-schema-spec/issues/1710 )

### Code of Conduct

- [x] I agree to follow this project's Code of Conduct

xfq:

A new blog post.

**Issue Number:**

-  Closes #2405

**Screenshots/videos:**

**Summary**

JSON is a very successful data format, but its string type is very simple. A value such as a title and name often needs more than Unicode characters. Consumers may need to know the language of the text and its direction, e.g., when displaying right-to-left scripts such as Arabic or Hebrew. Without that metadata, applications fall back to guessing, and guessing often fails.

The W3C Internationalization Working Group has a document on this topic: [Strings on the Web: Language and Direction Metadata](https://www.w3.org/TR/string-meta/), which is a practical guide for specifications and formats that exchange strings. This is a blog post about the best practice.

No.

# Checklist


vtushar06:
Hey @Utkarsh-123github, what do you think about this post?

pages/blog/posts/making-json-strings-work-for-every-language.md
```diff
@@ -0,0 +1,112 @@
...
+    "value": "HTML و CSS: تصميم و إنشاء مواقع الويب",
+    "lang": "ar",
+    "dir": "rtl"
+  }
+}
+```
+
+This gives producers and consumers a shared agreement. The producer can preserve the intended language and direction. The consumer can display the string correctly, apply the right `lang` and `dir` attributes in HTML, choose better fonts, select an appropriate text-to-speech voice, improve search and indexing, and avoid bidirectional text spillover.
+
+A schema can instead make the expected structure explicit:
```
gregsdennis: "instead" feels like the wrong word here.  Instead of what?  Maybe "also" can work, or just omit it?

pages/blog/posts/making-json-strings-work-for-every-language.md
```diff
@@ -0,0 +1,112 @@
...
+  "description": "A short description.",
+  "alternateTitle": {
+    "value": "جامعة المثال",
+    "lang": "ar",
+    "dir": "rtl"
+  }
+}
+```
+
+The document recommends `language` for a resource-wide default language field and `direction` for a resource-wide default direction field. For localizable text objects, the examples use `lang` and `dir`, which align naturally with HTML attributes when the value is eventually displayed.
```
gregsdennis, line 132: JSON and HTML don't need to be kept in alignment.  If the recommendation is for `language` and `direction`, it might be better to keep with that rather than to introduce an unnecessary association.

If the HTML/JSON alignment is warranted, an aside explaining why would be good.

pages/blog/posts/making-json-strings-work-for-every-language.md
```diff
@@ -0,0 +1,112 @@
...
+
+```json
+{
+  "title": {
+    "value": "HTML و CSS: تصميم و إنشاء مواقع الويب",
+    "lang": "ar",
+    "dir": "rtl"
+  }
+}
+```
```
gregsdennis, line 40: Maybe `title` might be the wrong property to use as an example.  When I saw this, I initially thought this was about changing the format of the `title` JSON Schema keyword rather than an instance property.

gregsdennis (commented):
I don't think there's any really wrong with the post or its content.  Just a few minor editorial suggestions.

I do have a concern that this is only tangentially related to JSON Schema, though.

pages/blog/posts/making-json-strings-work-for-every-language.md
jdesrosiers: Agreed. I had the same confusion in the original discussion. I think it would communicate the concept more clearly if a different property name was used.

jdesrosiers (commented):
> I do have a concern that this is only tangentially related to JSON Schema, though.

I see this as promoting best practices for schema authors, which I think is a great thing to include in this blog. But, I'd like to see a little more in the post giving guidance on how to apply these concepts using JSON Schema.

There are three patterns described in this post and you've only given a schema for one of them. I'd like to see schemas for all three patterns and examples of how to use them in a data structure the schema author is designing.

For example, the localizable text schema,

```json
{
  "$schema": "https://json-schema.org/draft/2020-12/schema",
  "$id": "https://exampe.com/localizable-text",

  "type": "object",
  "properties": {
    "value": { "type": "string" },
    "lang": { "type": "string" },
    "dir": { "enum": ["ltr", "rtl", "auto"] }
  },
  "required": ["value"]
}
```

And an example of using it,

```json
{
  "type": "object",
  "properties": {
    "title": { "$ref": "https://example.com/localizable-text" }
  }
}
```

Then the global settings pattern,

```json
{
  "$schema": "https://json-schema.org/draft/2020-12/schema",
  "$id": "https://exampe.com/global-localization",

  "type": "object",
  "properties": {
    "language": { "type": "string" },
    "direction": { "enum": ["ltr", "rtl", "auto"] }
  }
}
```

And an example of using it,

```json
{
  "$ref": "https://example.com/global-localization",
  "properties": {
    "name": { "type": "string" },
    "description": { "type": "string" },
    "alternateTitle": { "$ref": "https://example.com/localizable-text" }
  }
}
```

Then the language map pattern,

```json
{
  "$schema": "https://json-schema.org/draft/2020-12/schema",
  "$id": "https://example.com/language-map",

  "type": "object",
  "patternProperties": {
    "[a-z][a-z]": { "$ref": "https://example.com/localizable-text" }
  }
}
```

And an example of using it,

```json
{
  "type": "object",
  "properties": {
    "title": { "$ref": "https://example.com/language-map" }
  }
}
```

I think incorporating something like that would make this a more useful resource for schema authors.

jdesrosiers:
@jviotti I think there are several reusable structures here that could be a useful addition to your schema registry.

pages/blog/posts/making-json-strings-work-for-every-language.md
xfq: Filed https://github.com/w3c/string-meta/issues/94

pages/blog/posts/making-json-strings-work-for-every-language.md
xfq: fixed

xfq:
Thanks for the suggestions! I updated the blog post.

jdesrosiers (approved):
The updates look great. Thanks.

jdesrosiers:
@json-schema-org/web-team Please do your usual checks and merge when ready.

gregsdennis (approved):

